PR #1, opened 2020-07-14
tinkerbell proposals

Cbkhare:
## Description

Proposal for the tinkerbell enhancement. 

## Why is this needed

solves inconsistency, security and debugging isses.


DailyAlice (commented):
I am at least a big fan of the `tink-cli` proposal, as it's something that I have run into just when playing around testing the cli functionality for the docs. The docker container execution limitations are not insurmountable or un-documentable, but being able to run it separately would improve the experience.

mmlb:
Can we make this into just one file please?

displague:
@Cbkhare @mmlb 

@gianarb is figuring out a format like https://oxide.computer/blog/rfd-1-requests-for-discussion/ that will allow us to offer proposals and designs, such as the ones in this PR and the RFC's already present in the wiki, https://github.com/tinkerbell/tink/wiki/RFC-%233-Kubernetes-Support.  This repository may be renamed along the way.

Let's hold on this PR until that process is in place.

Cbkhare:
this PR needs change as per latest recommendation in RFC creation. I will update this PR or create will new.

displague:
Thanks for formalizing these proposals in the new format, @Cbkhare 🥇 . I expect we will close this issue once the HA proposal has been opened in a new PR.

I appreciate the goals and steps that you added between these initial proposals and the newly opened proposals.

You'll notice that I've provided some feedback on those proposals. While our proposal format is new, you have the lucky position of providing some of the most formative proposals.  As such, I'm asserting my interests in how they should be focused. 

A common theme in my comments is that I would like proposals to not restrict user choice when providing value.  Recommending or advising on particular use-cases is fair, but proposals should, IMHO, try to define the user value and provide the technical features of a solution, possibly exploring alternatives, without focusing on the implementation details.

The implementation details can be debated in PRs that apply the proposal in individual microservice repositories.

Cbkhare:
@displague , Thanks for the feedback. 

Yes, your comments are bringing a fresh perspective in RFD. I will modify the sections in all the RFD's with more emphasis on user needs and other possible solutions.

PR #2, opened 2020-07-29
Initial proposal about how to write Proposals

gianarb:


thebsdbox:
Looks good to me, I like the script to help generate the templates.. I'd like to make sure that the barrier to entry isn't too high that it scares away new contributors to the project also. Is there additional methods (more scripts or templates, or an instructional video etc..) that a new person would benefit from to use this process?

gianarb:
@thebsdbox ideally this is the place where we agree on the process and from there we will have to figure out all the work that has to be done in order to move this RFD into a `committed` state. One will be as you suggested:

- [ ] write READE.md for this repo explaining steps required to write a CFD copy-pasting in there what is needed in order to make the process super easy and clear for the community

Another will be for example:

- [ ] Set up a code generated site called cfd.tinkerbell.org that lists the status of the CFD

Will convert them to issue as soon as more people will send feedback. Does it sound good?

thebsdbox:
This sounds good to me. 👍🏻

0001/README.md
```diff
@@ -0,0 +1,396 @@
...
+* `Authors`: in the format of a list of `Name <email@emai.com>`
+
+Here a complete example:
+
+```yaml
+---
+id: 1
+title: Call for discussion duty
+status: ideation
+authors: \
```
displague: There is some discrepancy between `authors` in this example, `author` in the [metadata for this CFD](https://github.com/tinkerbell/proposals/pull/2/files#diff-f24504494b88bfe6c96cd7617c38ce08R5), and `authors` (not a yaml array) [below](https://github.com/tinkerbell/proposals/pull/2/files#diff-f24504494b88bfe6c96cd7617c38ce08R109).

0001/README.md
```diff
@@ -0,0 +1,396 @@
...
+Computer](https://oxide.computer/blog/rfd-1-requests-for-discussion/#rfd-life-cycle)
+looks familiar and functional for our purpose. They wrote the lifecycle in
+detail and for now we will follow it as it is.
+
+What follows here is copy/pasted from ["RFD 1 Requests for
+Discussion"](https://oxide.computer/blog/rfd-1-requests-for-discussion/) and all
+the credits goes to [Oxide Computer](https://oxide.computer).
+
+
+###RFD Metadata and State
```
displague: `RFD` vs `CFD`, this should also have a space after `###`.
https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md018---no-space-after-hash-on-atx-style-heading

gianarb: I am the best when it comes to confusing myself 😕 ❓

displague: I see.. This is how we are embedding their post. I wonder if @jessfraz could share the reuse license of this blog post, or perhaps knows another place on Github where this lives.

gianarb: I sent her an email 2 days ago. So, not sure yet but for we can keep the discussion going. Btw I am just confused. let's call it RFD because that's what they use :) Changing it now

0001/README.md
```diff
@@ -0,0 +1,398 @@
+---
+id: 0001
+title: Request for discussion HOWTO
+status: discussion
+pr: 2
```
displague: `pr: 2` This can not be known when creating PR because of race conditions. I don't see value in the numeric index value here.  On the other hand, permalinks to the PR and the Github file URL (in the default branch) would be useful.  These pieces of metadata could be added in commits or PRs that transition the `status` to later stages.  I don't think we need to try to capture them in the initial PR.

0001/README.md
```diff
@@ -0,0 +1,398 @@
...
+master), it is possible that the RFD is currently being created; stop and check
+with co-workers before proceeding! Once you have verified that the branch
+doesn't exist, create it locally and switch to it:
+
+```sh
+$ git checkout -b 0042
+```
+
+### CREATE A PLACEHOLDER RFD
+Now create a placeholder RFD. You can do so with the following commands:
```
displague, line 209: This is a nitpick, but the spacing around this header is not consistent.
https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md022---headings-should-be-surrounded-by-blank-lines

gianarb: this is solved, not sure why GitHub does not recognize it

0001/README.md
```diff
@@ -0,0 +1,400 @@
...
+* `Status`: description of the current state for the RFD
+* `Authors`: in the format of a list of `Name <email@emai.com>`
+
+Here is a complete example:
+
+```yaml
+---
+id: 0001
+title: Request for discussion HOWTO
+pr: 2
```
displague, line 85: This was missed when `pr: 2` was removed elsewhere.

0001/README.md
```diff
@@ -0,0 +1,400 @@
...
+authors: Gianluca Arbezzano <gianarb92@gmail.com>
+---
+```
+
+### Lifecycle
+
+I don’t want to rewrite the lifecycle here because the one in use by [Oxide
+Computer](https://oxide.computer/blog/rfd-1-requests-for-discussion/#rfd-life-cycle)
+looks familiar and is functional for our purpose. They wrote the lifecycle in
+detail and, for now, we will follow it as it is.
```
displague, line 96: "follow it as is."

0001/README.md
displague: https://twitter.com/jessfraz/status/1288473314713694209

displague (approved):
LGTM - This is a good start.  

I left some nitpicks.

0001/README.md
```diff
@@ -0,0 +1,400 @@
...
+the credit goes to [Oxide Computer](https://oxide.computer).
+
+~~~markdown
+
+### RFD Metadata and State
+
+At the start of every RFD document, we'd like to include a brief amount of metadata. The metadata format is based on the python-markdown2 metadata format. It'd look like:
+
+---
+authors: Andy Smith <andy@example.computer>, Neal Jones <neal@example.computer>
```
nathangoulding: I'd suggest we use at least one predominantly female name here, or gender neutral / placeholder names instead of 2 predominantly masculine.

nathangoulding (changes requested):

gianarb:
Thank you all for your feedback.
The proposal number 0001 is ready to go. It is in published state, it means that it can be modified via PR. The status will change to committed as soon as we will have a static site generated version of this REPO available to proposals.tinkerbell.org

PR #5, opened 2020-08-04
 0002: Set the mood around releasing code

gianarb:
I am actively looking for co-authors for this proposal.

### DIFF
--- a/0002/README.md
+++ b/0002/README.md
@@ -1,0 +1,196 @@
+---
+id: 0002
+title: Set the mood around releasing code
+status: accepted
+authors: Gianluca Arbezzano <gianarb92@gmail.com>
+---
+
+## Summary
+
+This proposal sets the expectation about what it means to release code part of
+the Tinkerbell organisation.
+
+## Goals and non-Goals
+
+Goals are:
+
+1. Define what a release is and what it is comprised of a release
+2. Define standard policies and practices for releases that all Tinkerbell
+   projects should follow
+3. Define Quality Assurance polices for releases.
+
+Non-Goals:
+
+1. How each project should be released and what it should release
+2. Define who own releases
+4. Define a roadmap and how it should be presented to the community.
+5. Support and deprecation lifecycle
+
+## Content
+
+A release is a tested and stable point in time snapshot from a codebase.
+
+We follow the [Semantic Versioning (semver) 2.0.0](https://semver.org).
+Summarized as follows:
+
+Given a version number MAJOR.MINOR.PATCH, increment the:
+
+* MAJOR version when you make incompatible API changes,
+* MINOR version when you add functionality in a backwards compatible manner, and
+* PATCH version when you make backwards compatible bug fixes.
+
+Additional labels for pre-release and build metadata are available as extensions
+to the MAJOR.MINOR.PATCH format.
+
+### Schedule
+
+Every subproject should be able to follow its own cadence in terms of how often
+or regularly a new release should be cut, as long as it is documented and
+shared with the community. It is a responsibility for the maintainer to figure
+out a good strategy that better fits their workload, process and codebase.
+
+`MAJOR` release right now are not even evaluated because the project is too
+young. By definition breaking changes require a new `MAJOR` release, right now
+any of our projects have a release yet, we are working hard to reach `v1.0.0`,
+until there bc breaks are allowed but they will be notified with a proper label
+`bc-break` attached to the pull request.
+
+A new `MINOR` version for Tinkerbell should be released following a proper
+schedule, right now I think every two months sounds reasonable looking at the
+code written and merged in the last couple of weeks. In this way the
+subprojects will have an intrinsic release target if they want to get their new
+code included in the Tinkerbell release.
+
+`PATCH` release for Tinkerbell can be done as required by the subprojects
+bug fixes.
+
+### GitHub release and release website
+
+Git supports tag and GitHub has the concept of a release based on a particular
+tag.
+
+The artifact generated as part of the release will be published as part of the
+release website, using as a guidance how [HashiCorp
+Release](https://releases.hashicorp.com/) works.
+
+I like the idea to use a release website vs having to push them as manifest in
+GitHub release because it gives a single place where the community can
+programmatically lookup what they are looking for, without having to look for
+repositories, versions and so on.
+
+As part of the Release page on GitHub we will publish a generated changelog.
+
+### Distribution
+
+Docker containers, binaries, packages such as `deb`, `rpm` requires
+infrastructure such as:
+
+1. A place where the binaries should be archived and easy do be downloaded from
+   (bintray, s3)
+2. Docker containers requires a registry
+
+In order to gain consistency across projects and to build familiarity for the
+community we will start as follow:
+
+1. Docker containers will be pushed to `quay.io` under the `tinkerbell`
+   organisation (this is already done)
+2. Binaries and possibly `rpm`, `deb` will be pushed to
+   [Bintray](http://bintray.com/) because it has a friendly open source plan.
+
+Right now there is not a formal group or structure that manages the Tinkerbell
+community needs in terms of infrastructure, servers and SaaS. We are building
+this group but nothing is formalized yet. As soon as the group will be assembled
+they will be the right target for questions, and requests about bintray, s3,
+Packet devices. In the meantime you can ask on
+[Slack](https://tinkerbell.org/community/slack/) in the #tinkerbell channel.
+
+### Architectures and operating systems
+
+The majority of the projects in the tinkerbell organisation right now are in Go,
+it means that it is easy to compile targeting different architectures and
+operating system. We also have an application in JavaScript (portal) and an
+operating system distribution (osie). It is not part of the this proposal to
+decide what Tinkerbell should support or not. But as a direction all the project
+should setup their release workflow in a way that will support:
+
+1. Multiple operating systems and architectures
+2. Static binaries and artifacts
+3. Docker containers
+
+### Ownership
+
+The contributors and maintainers for each project have the responsibility for
+the quality of the release itself.
+
+Tinkerbell as a project will have its group of interest around release
+management that will take care of it (this is out of scope for this proposal but
+has to be discussed in another one).
+
+### Quality Assurance
+
+Every project should be covered by tests: unit tests, functional tests and
+integration as needed.
+
+If all the test suite succeed the project can be released, otherwise tests
+should be fixed first.
+
+The documentation contains in tinkerbell.org covers the entire project and it
+has to be reviewed before any Tinkerbell release. Each Tinkerbell MINOR release will
+have its own documentation and users need a way to visualize the version they
+are looking for.
+
+### Tinkerbell
+
+Tinkerbell itself is more like a meta-project that glues all the subprojects it
+depends on. In terms of quality assurance we should have a strong and reliable
+pipeline of E2E tests to guarantee that the new version for the subprojects are
+working fine in integration.
+
+The subproject Tinkerbell depends on will be pinned as part of the Tinkerbell
+release but their version are a "suggestion". SemVer guarantees that all the
+PATCH release are working as long as the MINOR release is the same.
+
+Tinkerbell `v0.4.0` depends from boot `v1.2.0` every PATCH release for boot
+should work.
+
+SemVer guarantees that any MINOR release after the one pinned from Tinkerbell
+should work as well.
+
+Tinkerbell `v0.4.0` depends from boot `v1.2.0` every MINOR release for
+boot should work with Tinkerbell `v0.4.0` for example:
+
+* boot `v1.1.0` is not guaranteed to work
+* boot `v1.2.0` should work
+* boot `v1.2.4` should work
+* boot `v1.23.0` should work.
+* boot `v2.0.0` is not guaranteed to work
+
+### Continuous Delivery
+
+All subprojects should have a continuous delivery pipeline in place capable of
+creating a release from a tag.
+
+There should also be a well documented manual release process for the purposes
+of local testing and troubleshooting.
+
+### Documentation
+
+The documentation should be versioned and it should follow the same release
+cadence as Tinkerbell. It should be possible to jump between old version of the
+Tinkerbell documentation. How many, and how is out of scope for this proposal.
+
+## System-context-diagram
+
+## APIs
+
+## Alternatives
+
+I didn't find a well explained release process that can be reused. Every project
+has its own way even if, at the end the outcome is similar. I am listing here
+some successful release process I am involved with, please add yours if
+you have because I am sure this proposal will be a glue of the best experiences
+we have as a team:
+
+* [Kubernetes Release](https://github.com/kubernetes/sig-release/tree/master/release-team)
+* [Docker release page on GitHub](https://github.com/docker/docker-ce/releases)
+* [Kubernetes Release page on GitHub](https://github.com/kubernetes/kubernetes/releases)
--- a/prototypes/prototype.md
+++ b/prototypes/prototype.md
@@ -1,6 +1,6 @@
 ---
-id: 9999
-title: Place the RFC title here
+id: ID
+title: TITLE
 status: ideation
 authors: Name <email@email.com>
 ---
--- a/scripts/new.sh
+++ b/scripts/new.sh
@@ -2,17 +2,17 @@
 
 set -eu
 
-RFD_ID=$1
-RFD_TITLE=$2
+ID=$1
+TITLE=$2
 
-echo "🐙 >>> started the creation process for RFD ${RFD_ID} titled \"${RFD_TITLE}\""
+echo "🐙 >>> started the creation process for RFD ${ID} titled \"${TITLE}\""
 
-branch_exists=$(git branch -rl *${RFD_ID})
+branch_exists=$(git branch -rl *${ID})
 
 if [[ -z ${branch_exists} ]]; then
-    git checkout -b ${RFD_ID}
-    mkdir $RFD_ID
-    cp ../prototypes/prototype.md $RFD_ID/README.md
+    git checkout -b ${ID}
+    mkdir $ID
+    cat ./prototypes/prototype.md | sed s/ID/$ID/ | sed s/TITLE/"$TITLE"/ > $ID/README.md
     echo "🍀 >>> You are all set! Good luck"
     exit 0
 else


0002/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+Not-Goals:
+
+1. How each project should be released
+2. Define who own releases in practice
+3. Distribution via `apt`, `apk` and so on.
+
+## Content
+
+A release is a tested and stable point in time snapshot from a codebase.
```
DailyAlice, line 31: To keep to the scope of this I think just making it, "A release is a tested, documented, and stable point in time snapshot from a codebase." should cover versioning the docs.

0002/README.md
```diff
@@ -0,0 +1,164 @@
...
+has to be discussed).
+
+### Quality assertion
+
+Every project should be covered by tests: unit tests, functional tests and
+integration as needed.
+
+If all the test suite succeed the project can be released, otherwise tests
+should be fixed first.
+
```
DailyAlice, line 136: Include a review of the documentation (or at least any changes to the documentation) so that it accurately reflects the current features and functionality of the release.

0002/README.md
gianarb: Solved

0002/README.md
gianarb: Solved

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+A new `MINOR` version for Tinkerbell should be released following a proper
+schedule, right now I think every two months sounds reasonable looking at the
+code written and merged in the last couple of weeks. In this way the
+subprojects will have an intrinsic release target if they want to get their new
+code included in the Tinkerbell release.
+
+`PATCH` release for Tinkerbell can be done as required by the subprojects
+bug fixes.
+
+`MAJOR` release right now are not even evaluated because the project is too young.
```
displague: This is fair. If we had a roadmap we might populate it with the features that are needed to reach 1.0.0.
https://github.com/tinkerbell/tink/pull/241 😈

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+Additional labels for pre-release and build metadata are available as extensions
+to the MAJOR.MINOR.PATCH format.
+
+### Schedule
+
+Every subproject should be able to follow its own cadence in terms of how often
+or regularly a new release should be cut, as long as it is documented and
+shared with the community.
+
+A new `MINOR` version for Tinkerbell should be released following a proper
```
displague, line 58: I think it is fair to use this two-three month release windows to bump the minor version due to a wealth of added features.  Until we reach 1.0.0, we may want to bump the minor version whenever we are forced to introduce breaking changes that would have otherwise been patch releases. Is that obvious or should it be pointed out?

0002/README.md
displague: Major should be described before Minor, to match the earlier bullets.

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+### Architectures and operating systems
+
+We know there are different architectures and Go supports cross compilations.
+When possible we should cross compile binaries at least for:
+
+* Linux
+* macOS
+* Windows
+
+For `arm` and `x86_64`.
```
displague: `armv6`, `armv7`, `arm64`

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+boot should work with Tinkerbell `v0.4.0` for example:
+
+* boot `v1.1.0` is not guaranteed to work
+* boot `v1.2.0` should work
+* boot `v1.2.4` should work
+* boot `v1.23.0` should work.
+
+### Support lifecycle
+
+TODO: I need some help here, personally I am thinking about 1 year (6 Tinkerbell
```
displague: Can this be explicitly deferred until v1?

gianarb: Yes! I was desperately looking for what to kick out from this proposal! I will move it to a non-goal!

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+status: ideation
+authors: Gianluca Arbezzano <gianarb92@gmail.com>
+---
+
+## Summary
+
+This proposal sets the expectation about what it means to release code part of
+the Tinkerbell organisation.
+
+## Goals and not Goals
```
detiber: nit
```suggestion
## Goals and Non-Goals
```

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+Goals are:
+
+1. What is a release
+2. Set a few rules that all the projects should follow
+3. Quality assertion.
+
+Not-Goals:
```
detiber: nit:
```suggestion
Non-Goals:
```

0002/README.md
detiber: Should this also call out what types of artifacts we expect for a release?

For example:
- archive of the source code at the time of release
- binary build artifact(s) for a given release
- container image(s) for a given release
- OS specific packages for a given release
- etc

0002/README.md
detiber: +1 I think this should likely call out the possibility of breaking changes landing in a minor release prior to v1.0

0002/README.md
detiber: +1 to linking to an umbrella issue, milestone, or some other way of tracking issues needed to reach 1.0

0002/README.md
detiber: Should this be `arm64` here?

Or are there also plans to support 32 bit arm?
- If so, which arm target for linux (`armhf` vs `armel`)? I'd assume `armhf`
- Which GOARM version? I'd assume `GOARM=7`

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+This proposal sets the expectation about what it means to release code part of
+the Tinkerbell organisation.
+
+## Goals and not Goals
+
+Goals are:
+
+1. What is a release
+2. Set a few rules that all the projects should follow
+3. Quality assertion.
```
detiber:
```suggestion
3. Define Quality Assurance polices for releases.
```

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+## Summary
+
+This proposal sets the expectation about what it means to release code part of
+the Tinkerbell organisation.
+
+## Goals and not Goals
+
+Goals are:
+
+1. What is a release
```
detiber:
```suggestion
1. Define what a release is and what it is comprised of
```

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+2. Set a few rules that all the projects should follow
```
detiber:
```suggestion
2. Define standard policies and practices for releases that all Tinkerbell projects should follow
```

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+tag.
+
+The artifact generated as part of the release will be published as part of the
+release website, using as a guidance how [HashiCorp
+Release](https://releases.hashicorp.com/) works.
+
+I like the idea to use a release website vs having to push them as manifest in
+GitHub release because it gives a single place where the community can
+programmatically lookup what they are looking for, without having to look for
+repositories, versions and so on.
```
detiber, line 79: We could leverage something like goreleaser (at least for go projects) to automate the release process using github releases, and we can leverage the `latest` release link for the common use cases.

That might make a decent solution, at least for individual subprojects.

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+### Ownership
+
+The contributors and maintainers for each project have the responsibility for
+the quality of the release itself.
+
+Tinkerbell as a project will have its group of interest around release
+management that will take care of it (this is out of scope for this proposal but
+has to be discussed in another one).
+
+### Quality assertion
```
detiber:
```suggestion
### Quality Assurance
```

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+The documentation contains in tinkerbell.org covers the entire project and it
+has to be reviewed before any Tinkerbell release. Each Tinkerbell MINOR release will
+have its own documentation and users need a way to visualize the version they
+are looking for.
+
+### Tinkerbell
+
+Tinkerbell itself is more like a meta-project that glues all the subprojects it
+depends on. In terms of quality assertion we should have a strong and reliable
```
detiber:
```suggestion
depends on. In terms of quality assurance we should have a strong and reliable
```

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+* boot `v1.1.0` is not guaranteed to work
+* boot `v1.2.0` should work
+* boot `v1.2.4` should work
+* boot `v1.23.0` should work.
+
+### Support lifecycle
+
+TODO: I need some help here, personally I am thinking about 1 year (6 Tinkerbell
+releases)
```
detiber: I think it might be a bit early to define the support lifecycle just yet at least until v1.0, maybe one of the blocking criteria for v1.0 should be that there is a defined support lifecycle.

For pre-1.0, I think it should be sufficient to say that only the current (non-pre-release) minor release is supported.

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Continuous Delivery
+
+All the project should have in place a continuous delivery pipeline capable of
```
detiber:
```suggestion
All subprojects should have a continuous delivery pipeline in place capable of
```

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+creating a release from a tag.
+
+The process needs to be documented with the possibility to run it locally for
+testing purpose and for troubleshooting.
```
detiber:
```suggestion
There should also be a well documented manual release process for the purposes of local testing and troubleshooting.
```

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Documentation
+
+The documentation is versioned as well following the release cadence Tinkerbell
+has. A new version of Tinkerbell goes out when E2E tests are green and
```
detiber:
```suggestion
has. A new version of Tinkerbell goes out only when E2E tests are green and
```

0002/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+All the project should have in place a continuous delivery pipeline capable of
+creating a release from a tag.
+
+The process needs to be documented with the possibility to run it locally for
+testing purpose and for troubleshooting.
+
+### Documentation
+
+The documentation is versioned as well following the release cadence Tinkerbell
```
detiber:
```suggestion
The documentation should be versioned and follow the same release cadence as Tinkerbell
```

0002/README.md
gianarb: I think this proposal should stay out of what each project exposes. The majority of the project is in Go, Osie is an OS distribution and it follows different rules. As I tried to explain but I will reinforce this is a generic high-level way approach to releasing software that every sub-project has to follow and apply.

I am sure we will have a follow-up issue/proposal or who knows about what Tinkerbell as a project will support in terms of OS, architectures, distribution, and so on and by consequence, the underline projects will have to get inline to those requirements.

0002/README.md
gianarb: ROADMAP is out of scope from this proposal but I am happy to read one about it.
I am saying that because I am not even sure today about how far we are from 1.0.0 or 0.0.1 or 0.1.0

0002/README.md
displague: Osie results in a binary (executable at bootup, which tftp users would consume), does it not?
In Tinkerbell, is osie ever consumed as a container (maybe as an OCI filesystem / glorified-tarball)?

0002/README.md
displague: fwiw, I **really** think we should stamp 0.1.0 where we are and worry about where we are going later. We need a relative starting point (tinkerbell/tink#74). 

0.1.0 is preferred to 0.0.1, https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

0002/README.md
gianarb: It is a tarball and maybe it will become a set of tarball who knows right now. It can be confused as a docker image I am sure but don't know if we do it

0002/README.md
gianarb: Done

0002/README.md
gianarb: Yes, the problem with goreleaser is that it does not work great when it comes to multi arch support for Docker containers

displague: Some excellent resources here:
* https://github.com/goreleaser/goreleaser/issues/530
* https://www.youtube.com/watch?v=xdehEUG1RFY 😄

0002/README.md
gianarb: This proposal is about how we should do releases. Asking to tag 0.1.0 where we are looks unreasonable to me. We don't know where we are and we are not sure about where to go at this phase. The coordination required to get this kind of information for all the projects looks out of scope and not that easy.

Even if we tag a release today we don't have the right setup to use that in docker-compose or in other places and the binaries won't be generated. So I think it is a waste of time and I prefer to make it better.

I am happy to repeat myself even if it looks out of scope to me but if you are curious before cutting any release I would like to get:

1. A way to consume a pinned version of our tools where it is needed (setup.sh, docker-compose)
2. Check what automation does when a release is cut, tink does not do anything as it is today
3. Having maintainers for each project that can figure out the release needs and achieve them by themself. 

can you tell me why you think you need a release? During a chat with @thebsdbox  I wrote this proposal that I think will solve an actual issue without requiring a release https://github.com/tinkerbell/proposals/pull/11

thebsdbox: I agree with @displague that we need to get something out there that is "stable" enough for the end-user, as they need something to get their hands on that "just-works™". As development continues on Tinkerbell we've broken various components multiple times and left end-users with no way to know what they're running or anyway of rolling-back. We're looking at pinning a set of `image:sha` until all of the micro services have maintainers and releases.

gianarb: btw I agree with you as well, I just think it is out of scope for this proposal :) now we have #11 that we can use to discuss that and I am happy enough if @displague and you (@thebsdbox) are happy as well!

0002/README.md
displague: Am I wrong to think that maintaining an independent release website is going to be a burden on project maintainers?

If this is just object storage (url+key) as you suggest below as an option for distribution we may not be able to get it for free, but I think maintenance will be less cumbersome than bintray.

> using as a guidance how [HashiCorp Release](https://releases.hashicorp.com/) works

The Hashicorp example is interesting because they have moved away from releases.hashicorp.com for Terraform providers (now contained in a registry).

> I like the idea to use a release website vs having to push them as manifest in GitHub release because it gives a single place where the community can programmatically lookup what they are looking for

The GitHub releases can be accessed through the [GitHub API](https://developer.github.com/v3/repos/releases/#list-releases (`GET)
) very easily:
```sh
$ curl https://api.github.com/repos/tinkerbell/tink/releases
[

]
```

> without having to look for repositories, versions and so on

I think this can be solved through the GitHub `latest` keyword being used in URLs in-place of a version, and by distributing our binaries through other systems (docker, debian, fedora, etc) where users won't need to worry about versions (maintainers will).

displague: I'm wondering if using Github releases just saves us time and effort - it's a solved problem with no immediate overhead.

0002/README.md
```diff
@@ -0,0 +1,197 @@
...
+I like the idea to use a release website vs having to push them as manifest in
+GitHub release because it gives a single place where the community can
+programmatically lookup what they are looking for, without having to look for
+repositories, versions and so on.
+
+As part of the Release page on GitHub we will publish a generated changelog.
+
+### Distribution
+
+Docker containers, binaries, packages such as `deb`, `rpm` requires
```
displague, line 85: We can publish debs to a PPM via launchpad.net (ubuntu/debian official path): https://help.launchpad.net/Packaging/PPA

I don't know about rpm build/archive sites, but a quick Google turns up: https://build.opensuse.org/ and https://fedoraproject.org/wiki/Koji

We could defer these types of archives for a later proposal (or we could treat them like simple binaries and stash them with each github release).

0002/README.md
displague: I recall the interest in nightly builds that are not based on tags or releases. I don't see that stated in this proposal (perhaps it is in another). If that thought is influencing this decision, I'd caution that this, as a requirement, may be causing us more grief than it provides in value.

displague: I'm wondering if we can simplify the organization's requirements, for now, and prevent the need for https://github.com/tinkerbell/proposals/pull/5/files#diff-5ae2d202d9cad195081db97be153eaa9R101-R106.

A maintenance team would certainly be needed to coordinate with interested vendors to prepare, build, certify, and deploy tink in their environments.  Do we have to figure this out now? Can this infrastructure maintenance team (and the additional infrastructure requirements) be the target of a future proposal?

0002/README.md
```diff
@@ -0,0 +1,197 @@
...
+it means that it is easy to compile targeting different architectures and
+operating system. We also have an application in JavaScript (portal) and an
+operating system distribution (osie). It is not part of the this proposal to
+decide what Tinkerbell should support or not. But as a direction all the project
+should setup their release workflow in a way that will support:
+
+1. Multiple operating systems and architectures
+2. Static binaries and artifacts
+3. Docker containers
+4. APT, DEB, RPM package manager.
```
displague: apt and deb are the same, like dnf to rpm.

Perhaps we should remove all deb/rpm requirements for now.

We are not prepared to build, host (could be on launchpad, koji, etc), or maintain (install/uninstall scripts, dependency requirements) these artifacts.

0002/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Ownership
+
+The contributors and maintainers for each project have the responsibility for
+the quality of the release itself.
+
+Tinkerbell as a project will have its group of interest around release
```
displague, line 125: If we can define the current `maintainers` as the "team" responsible for release management, for now, we don't have to defer this work to a future proposal. (and we can always add another proposal to redefine the release team or process).

0002/README.md
```diff
@@ -0,0 +1,197 @@
...
+### Quality Assurance
+
+Every project should be covered by tests: unit tests, functional tests and
+integration as needed.
+
+If all the test suite succeed the project can be released, otherwise tests
+should be fixed first.
+
+The documentation contains in tinkerbell.org covers the entire project and it
+has to be reviewed before any Tinkerbell release. Each Tinkerbell MINOR release will
```
displague, line 138: I see that this `MINOR` release requirement for documentation may be adding to your hesitance to tag everything `0.1.0`. I'm fine with `0.0.1` for what it's worth :-)

0002/README.md
```diff
@@ -0,0 +1,197 @@
...
+SemVer guarantees that any MINOR release after the one pinned from Tinkerbell
+should work as well.
+
+Tinkerbell `v0.4.0` depends from boot `v1.2.0` every MINOR release for
+boot should work with Tinkerbell `v0.4.0` for example:
+
+* boot `v1.1.0` is not guaranteed to work
+* boot `v1.2.0` should work
+* boot `v1.2.4` should work
+* boot `v1.23.0` should work.
```
displague, line 165: for completeness:
```
* boot `v2.0.0` is not guaranteed to work
```

0002/README.md
```diff
@@ -0,0 +1,197 @@
...
+creating a release from a tag.
+
+There should also be a well documented manual release process for the purposes
+of local testing and troubleshooting.
+
+### Documentation
+
+The documentation should be versioned and follow the same release cadence as
+Tinkerbell has. A new version of Tinkerbell goes out only when E2E tests are
+green and
```
displague: "green and ..."
Missing documentation. typical OSS project 🚀

0002/README.md
displague: The order of these paragraphs should be either "major", "minor", "patch", or possibly the reverse. I think it is confusing to lead with "minor".

0002/README.md
displague: @gianarb While you do clearly reflect this below, https://github.com/tinkerbell/proposals/pull/5/files#diff-5ae2d202d9cad195081db97be153eaa9R139-R142, I think @DailyAlice's comment is not yet reflected in this line:

> A release is a tested, **documented**, and stable point in time snapshot from a codebase.

0002/README.md
displague: I continued my rant in https://github.com/tinkerbell/proposals/pull/11#pullrequestreview-472114109 and then came back to this PR for another review. 

I'm looking forward to getting this merged soon. I've made some recommendations to pull more topics out-of-scope to assist that.

0002/README.md
gianarb: > If we can define the current maintainers as the "team" responsible for release management

Sometimes I think I leave in another world, do we have maintainers?! I don't see them  🙈  If you have a list can you share that split by repository? It will solve my next task

0002/README.md
gianarb: I don't want to tag anything as part of this proposal

thebsdbox:
I think with #11 in place we have enough coverage for both personas:
- People who develop on Tinkerbell can understand the code process
- People who use Tinkerbell have the beginnings of a simplified deployment process

I'm happy for this proposal to be merged as is.

PR #7, opened 2020-08-11
0003: ADD RFD for centralized logging in tinkerbell

Cbkhare:
## Description

Centralized logging for Tinkerbell


0003/README.md
```diff
@@ -0,0 +1,116 @@
...
+- By default, the logging feature would be supported by rsyslog. 
+- Logging feature should be implemented in such a way that a user can modify the type of logging (rsyslog or fluentbit) as per his needs. This may require the user to make additional changes on his end to enable that feature. 
+
+## Goals: 
+
+Goals cover the feature Details:
+- rsyslog is unix based package to capture the logs in a connected network. This entails having centralized logging using rsyslog. 
+- rsyslog is based on the top of syslog. 
+- rsyslog reads the log in syslog daemon service.
+- rsyslog will run on a logging server. Description of this logging server will be provided as a part of logging configuration.
```
displague: These goals include strong opinions about the logging agent. There are many choices available and depending on where Tinkerbell is run, rsyslog may not be the best choice. For example, systemd, syslogd, and others are the default in many environments.

Is the choice of logger also dictating the preferred operating system? The preferred operating environment?

If the Tinkerbell microservices are run from Kubernetes would we have the same expectation that a syslog client is used?

I would prefer to see the focus shift to more detail about the form and format of logging, for example (just an example):

- Tinkerbell microservices must include [OpenTracing](https://opentracing.io/specification/) spans. For example, in Go microservices <https://github.com/opentracing/opentracing-go#serializing-to-the-wire> and <https://www.jaegertracing.io/docs/1.18/architecture/#jaeger-client-libraries>.
- The Tinkerbell control-plane will support the syslog protocol https://tools.ietf.org/html/rfc5424 (as implemented by rsyslog) for the purposes of:
  - forwarding logs to another logging agent 
  - collecting microservice logs (optional)
- Tinkerbell microservice log messages come in the form "...."
  - Messages from the xyz service also include "..."

By reducing the statements about what role rsyslog may play, this draft for logging standards and collection unbinds itself on the limitations of a single logging tool. We can remove focus from specific install environments (docker-compose).

Most of the rsyslog context that has been provided in this part of the proposal could be retained and migrated to rsyslog_provisioner_configuration.md, including docker-compose, rsyslog configurations, and microservice configurations.

0003/README.md
```diff
@@ -0,0 +1,116 @@
...
+- Worker container Files to have workflow Id. In the current framework, this can be done. Changes would be required to attach id of workflow in the name of worker action containers. 
+- Files for each node will be placed in the respective directory on logging server.
+
+
+5) Implementation with fluentbit and other logging drivers:  
+* Separate doc on how rsyslog, fluentbit would work. 
+
+
+## Documentation:
+
```
DailyAlice, line 116: transitioning the content here to documentation should go pretty smoothly, this is an excellent write up.

0003/README.md
```diff
@@ -0,0 +1,116 @@
...
+This implementation of rsyslog has been verified with POC and been discussed in-depth within the Packet and Infracloud team.  
+
+## Suggestions:
+
+- We want to ensure at a basic minimum there is a centralized logging framework which exists for tinkerbell. In order to have a generic implementation of configuring log-drivers below are some proposals.
+
+1) How to make action containers run with logging drivers.
+
+a) Place the logging details inside the workflow template.
```
DailyAlice: From the perspective of user experience: I think that adding it to the template is easy to explain and surface in documentation. Which is not quite the same as the easiest or best for actual user experience, but just another perspective on getting it in front of users.

Cbkhare: @DailyAlice 

yes, having in the template is a good option. From the perspective of usage is where we are getting a mixed feeling. The template is more focused on workflow and logging is more focused on the infrastructure level, but again logging is for actions containers itself. So at one hand where it seems good, on the other hand, it seems like we should also look for some other options as well.

0003/README.md
Cbkhare: @displague 

With this RFD, we are proposing to centralize the collection of microservices logs of provisioner and worker. This system in the Tinkerbell will be a bare minimum framework which user can use as a part of day 0 configuration and beyond. Additionally, We want to give the user a choice to develop his own logging collection system on top of it, as per his needs.

RFD right now focuses on docker logs and log-driver becomes an important aspect of implementation. With types of log-driver available, best-suited ones are `journald`, `syslog` and `fluentbit`. Worker has alpine image which does not have journald, And fluentbit which was earlier present but later was discarded on the grounds of testing. Also, for fluentbit we need additional services like kibana and ELK. 
So, we concluded on having `syslog` as the log-driver and  `rsyslog` as the agent which internally depends on syslog itself. Alternatively, syslog-ng can be used, but I couldn't find much difference between these two agents. 

Since we are using syslog as the log-driver, platform dependency is only one that we need a Linux platform. As a part of the implementation, dependency will be there to have Syslog be installed on the Linux image. 

On Kubernetes front, the underlying logging agent will depend upon the log-driver of docker-container. 

Distributed tracing suggestion is definitely encouraging and we should have it in the Tinkerbell. When I started the POC, implementing it was my first thought. But in my understanding, I feel we should have a framework first on collecting docker logs. Also distributed tracing logs are going to be pushed onto docker container logs and can be collected in `ELK` which is a bit similar to our past implementation of `fluentbit`. If you think there is a way where we can use clients like `Jaeger` to have centralized logs without using `ELK` or any other collector, then please give me some links, I will explore those options as well. Tracing implementation may be out of the scope of this RFD but I can definitely add it in the future road map and get it done in next steps.

I agree with your suggestion of making this file more on user requirements and feature details, with separate files on rsyslog implementation. I will make these changes.

Cbkhare: @displague I have modified the README.md and added other required files.

Updated future roadmap with distributed tracing.

0003/README.md
```diff
@@ -0,0 +1,123 @@
+---
+id: 0003
+title: Centralized Logging for Tinkerbell microservices.
+status: POC Done, Code change ready
```
gianarb: The status is not one of the accepted status please have a loot at proposal 0001 and change it

0003/README.md
```diff
@@ -0,0 +1,123 @@
...
+---
+
+## Summary:
+
+Centralized logging framework for Tinkerbell microservices. 
+
+## Overview: 
+
+- Tinkerbell will have the feature of centralized logging. 
+- Logging feature should be implemented in such a way that a user can modify the type of logging (rsyslog or fluentbit) as per his needs. This may require the user to make additional changes on his end to enable that feature.
```
gianarb: The user should be able to modify the location where the log will be centralized. It does not need to be the provisioner

0003/README.md
```diff
@@ -0,0 +1,123 @@
...
+## Overview: 
+
+- Tinkerbell will have the feature of centralized logging. 
+- Logging feature should be implemented in such a way that a user can modify the type of logging (rsyslog or fluentbit) as per his needs. This may require the user to make additional changes on his end to enable that feature. 
+
+## Goals: 
+
+Goals cover the feature Details:
+- Tinkerbell will have the feature of centralized logging of microservices. 
+- Logging feature should be very lightweight i.e without additional microservices.
```
gianarb, line 21: Specify that lightweight is a requirement only for local/vagrant setup. In production or other use should be able to set up what they already have (if they use papertrail or logstash and so on), or what they want.

0003/README.md
gianarb: Should we have a "default" logging strategy to set at the `tink` layer with the possibility of overriding it as described?! my expectation is that 99% of the time logging will go to the same place

0003/README.md
gianarb: We need to cover those topics as part of the first release.

1 - We need to document a high level of how logging works, where it takes action, and what's the purpose and when how to use it.
2 - We should add as part of the Setup guides a reference to (1) + the concrete centralization logging strategy the setup uses. Vagrant will use rsyslog, so we will have to link rsyslog documentation to Vagrant setup.

gianarb:
@Cbkhare great start! I added a few comments and I am working on a way to render the proposals repo as a website, but in order to make this to work you have to include everything (other than the images that are already linked) as part of the README.md otherwise it won't get rendered

0003/README.md
gianarb: @displague the proposals describe well the goal and the fact that the logging centralization will be flexible as almost every other I used in the past is. But we need to start with one. And the Vagrant setup guide requires the smallest one we can come out with.

Tracing will have its own proposal, it is a big topic and I don't like the idea to mix them all together. Let's leave it out. (@Cbkhare can you specify that tracing is not part of this proposal as NON GOAL and remove reference of it?) Thanks

0003/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+Example with rsyslog
+
+![rsyslog](rsyslog.png)
+
+## Future roadmap:
+
+As a part of roadmap implementation, certain changes in the existing framework would be required to bring below proposals onboard.
+
+1) Implementation of Distribute tracing on the microservices.
```
gianarb: Let's leave tracing for another proposal

0003/README.md
Cbkhare: yes, ryslog will be the default logging structure. This point is more inclined to user needs if he is willing to modify the log-drives or log files location.

0003/README.md
Cbkhare: User will have the flexibility to modify the log driver and log server as well. By default, the logs will be aggregated to provisioner. I have mentioned this in `Content`

0003/README.md
Cbkhare: This will be a lightweight feature for any form for deployment, including Terraform. 

Changes required for this implementation will be at
- provisioner creation level.
- tink services (irrespective of the fact that we bring services up from docker-compose or another method. As of now I am aware of only docker-compose, but during implementation, I will embed this feature on all the forms of deployment)
- worker and its action containers. 

User will have the flexibility to build something on the top of that. It can be done by changing the log-driver and adding required microservice which will interact with those log drivers. We will have this in the documentation along with an example.

As of now, the implementation plan is to build a generic logger and then have a way for the user to build on the top of that.

0003/README.md
Cbkhare: I have added it in the road map. Should we remove it from there?

0003/README.md
Cbkhare: done

Cbkhare:
> @Cbkhare great start! I added a few comments and I am working on a way to render the proposals repo as a website, but in order to make this to work you have to include everything (other than the images that are already linked) as part of the README.md otherwise it won't get rendered

@gianarb , I didn't get this one. Are you talking about referencing other md files inside the README.md

0003/README.md
gianarb: I think we should not mention tracing at all in this proposal

gianarb:
> @gianarb , I didn't get this one. Are you talking about referencing other md files inside the README.md

yes. The `docker-compose.yaml` and the other `markdown` files won't be rendered in the static site generated version of this repository. You have to find a way to move that content as part of the main `README.md`.

Thanks

0003/README.md
gianarb: Can you double check!? I still see the old status

Cbkhare: I think while checking how to reference other Md files in readme.md, I think I missed pushing this change. :|

0003/README.md
Cbkhare: i have removed it for now. We can take it up as separate RFD.

0003/README.md
Cbkhare: I agree, as a part of this change document will cover
- Logging architecture to help the user understand our implementation.
- How rsyslog is used
- How other log-driver can be used.

In the first phase of implementation, we will add point 1 and 2. After Tinkerbell as a product is ready where a user can change log-driver, we will add point 3, maybe also with an example.

thebsdbox:
Just been through this, on the face of it and from all of the discussions that we've had Im happy with this proposal.

Cbkhare:
Merging the PR. Development will be taken forward with https://github.com/tinkerbell/osie/issues/58

PR #9, opened 2020-08-12
0004: ADD RFD for Tinkerbell Security

Cbkhare:
## Description

RFD for Tinkerbell security.


0004/README.md
```diff
@@ -0,0 +1,41 @@
...
+authors: chitrabasu khare <chitrabasukhare89@gmail.com>
+---
+
+## Summary:
+
+Currently, all the services in provisioner are accessible from outside (i.e via worker). Even in some implementation, ex tink #195, worker have full access (Read+write) tink-server via tink client. This leads to the possibility that worker can alter the behaviour of tink-server.
+
+## Goals:
+
+- Move all Tinkerbell services in private Network and expose them via proxy on a given port.
```
displague, line 14: I generally find that I am looking for more context in this proposal.  I'll use the goal as an example since the context I am looking for around this goal would have a strong influence on the content and perhaps the summary too.

> Move all Tinkerbell services in private Network

In what scenario can we assume that tinkerbell has not already been installed into private-only network?  Does this proposal recommend changes to the default installer options or the services in general?

I think this implies that we will be changing the installation defaults. We offer 3 different installers. Will we be using a common address in all of these environments (I assume not)? If not, what principle will be used to define the private network (options, preferred interface (eth1?), preferred rfc-1918 range (10.x) )?

I would like to see this proposal add more detail about how this goal could be achieved.
Defining a common set of microservice parameters, such as `--bind-address 0.0.0.0 --bind-port 1234` or `--bind-address ::1`, for example, allows the user to decide what makes a private or protected network.

> expose them via proxy on a given port

Likewise, this could be offered as a `--proxy-address` and `--proxy-port`.  Would we get the existing behavior without these parameters? Would the default `proxy-address` be the `bind-address`?  What services (specifically) would be proxied?


Documenting these parameters and instrumenting our installers to make private addresses the default (perhaps behind a toggle) would be secondary goals.

0004/README.md
```diff
@@ -0,0 +1,41 @@
+---
+id: 0004
+title: Tinkerbell Security
```
displague, line 3: I would expect a "Security" proposal to focus on the practices around security disclosures (reception and transmission) and the handling of the sensitive patches.

I think a better title for this would be "Network Interfaces and Visibility of Tinkerbell Services".

gianarb:
I think this suffers the same problem #10 has. The scope is too big, the title is too generic and it is hard to build a discussion around this proposal.

0004/README.md
```diff
@@ -0,0 +1,41 @@
...
+status: ideation
+authors: chitrabasu khare <chitrabasukhare89@gmail.com>
+---
+
+## Summary:
+
```
gianarb, line 9: I thinik this is a NO-PROBLEM they are accessible from the outside because we made a docker-compose that can be used in a very limited amount of scenarios. Usually for testing and to try the project out.

Cbkhare:
I will revert back on comments with more details in some time.

gianarb:

2. Merge it

---

### For the Author

PR #10, opened 2020-08-14
0006: ADD RFD for Resiliency in Tinkerbell Provisioner

Cbkhare:
## Description

RFD proposing Resiliency in Tinkerbell provisioner.


0006/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+## Goals:
+
+- Bring resiliency in Tinkerbell. 
+- Tinkerbell as a setup should be highly available.
+
+## Content: 
+
+- We would like to have a framework where Tinkerbell as a solution is fault-tolerant.
```
gianarb, line 35: Can you add a couple 5-6 different scenarios that you would like to fix in this proposal?

0006/README.md
```diff
@@ -0,0 +1,49 @@
...
+- All most all of the core services are managed by provisioner. So, it will be good to have a framework where provisioner is highly available.
```
gianarb: Can you be more precise? Why it is not "reliable" today? I would like to be sure that you are not confusing what you see in the "Setup Vagrant" and the "Setup Packet" and the docker-compose with the reliability of the Tinkerbell project because those are two different things.

0006/README.md
```diff
@@ -0,0 +1,49 @@
...
+- It can be a multinode provisioner solution with active-active or active-passive implementation.
```
gianarb: This is a good point, and maybe it can be a proposal by itself.. Even if I do not really understand it.

if you deploy Postgres in RDS or with its own full tolerant strategy you can have 100 tink-server behind a proxy already today.

0006/README.md
```diff
@@ -0,0 +1,49 @@
...
+- The Provisioner services should be partition tolerant during network failures or communication failure between provisioner nodes.
```
gianarb, line 50: This is a good proposal by its own

0006/README.md
```diff
@@ -0,0 +1,49 @@
...
+- Additional logic would be required to sync services running inside provisioner nodes.
```
gianarb: Which one? can you more detailed?

0006/README.md
```diff
@@ -0,0 +1,49 @@
...
+- A separate machine to place an L3/L7 Load balancer or proxy would be required.
```
gianarb: This is something that can be done today and it does not require any work. Vagrant setup is just a way to setup Tinkerbell, the same as the docker-compose. This to me looks more like something that has to be developed in the documentation and as ops experience.

gianarb (changes requested):
The title is too generic and it makes the scope of this proposal hard to discuss and understand.

my feeling is that you are confusing the work you see in docker-compose, vagrant setup with the Tinkerbell architecture itself.

A lot of the points you raised are tradeoff we had to make with docker-compose and Vagrant to provide to the community an easy way to try Tinkerbell.

Obviously it does not scale or it is not "full tolerant" but that's fine.

0006/README.md
detiber: Would it help if there was also a Helm chart, kustomize, or some other type of Kubernetes artifact available for deploying the services?

0006/README.md
detiber: +1, I think this is likely something that would make sense to address on a component-by-component basis, since the behavior we'd want to invoke may not be generalizable.

0006/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+## Problems and Suggestions:
+
+1) How does worker communicate with provisioner?
+-  The worker will communicate to the load balancer (or proxy) and the request will be routed to a replica of provisioner.
+-  Sessions need to be maintained inside Load balancer so that different requests of worker goes to the same replica of provisioner. 
+-  How sessions can be maintained with the latest changes of GRPC streams, requires some investigation.
```
detiber, line 60: Are "sticky" sessions needed for all services? Or only a subset of the services?

0006/README.md
```diff
@@ -0,0 +1,49 @@
...
+## Problems and Suggestions:
+
+1) How does worker communicate with provisioner?
+-  The worker will communicate to the load balancer (or proxy) and the request will be routed to a replica of provisioner.
+-  Sessions need to be maintained inside Load balancer so that different requests of worker goes to the same replica of provisioner. 
+-  How sessions can be maintained with the latest changes of GRPC streams, requires some investigation. 
+
+2) Which services require sync among it replicas?
+-  Currently, at least DB service needs a sync logic in backend. 
+-  This can be achieved by having daemon agents running in both nodes and syncing the database.
```
detiber, line 64: I kind of like the idea of externalizing the complexity of managing a replicated data store if we can. As @gianarb suggested, something like RDS would fit nicely for this and many folks are already capable of deploying or managing databases on their own. Using a non-replicated db for demo/trial purposes is fine, but I don't think we'd want to provide generalized community support for replicated DBs for production workloads.

0006/README.md
Cbkhare: It is sync of tink-db service, i.e how will db service will sync volumes they are mounted and keep the cluster consistent.

Cbkhare:
@gianarb I have gone through your comments. Let me see how I can refactor `README.md` more.

since https://github.com/tinkerbell/proposals/pull/7 is our priority right now, I will change this PR parallelly in this week or by mid next-week.

0006/README.md
Cbkhare: done

0006/README.md
Cbkhare: My suggestion is to have this feature for an active-active multi-node provisioner. In that case request needs to be routed to one of the node, hence we need a LB. It should be independent of the `type of deployment` of setup.

0006/README.md
Cbkhare: will push change in some time.

0006/README.md
Cbkhare: @gianarb I have updated the doc on the reliability part. I am proposing this solution which should be indepented on the form of deployment or setup type.  

@detiber yes, K8s can be one of the ways to do it. With K8s we can have multiple instances of the services plus scalability can also be managed by K8s itself. 
One challenge we could possibly face is (which this PR is also trying to address is) what is the master K8 node goes down 9due to network or hardware failure) in that case also Tinkerbell as setup will fail. Though, We can have multi-master K8s cluster to solve that problem.

0006/README.md
Cbkhare: Yes, this point demands a bit extensive implementation. but needs to be covered as a part of this proposal itself. 

IMO, we will have to create a number of subtask under this issue.

0006/README.md
Cbkhare: @detiber I agree with you. 

We can start with DB first and then scale onto other services.

0006/README.md
Cbkhare: As of now, most of micro-services are exposed to the worker which it uses for workflow execution, like boots, tik-server etc. In that case, session needs to be maintained by at least those services. 
I am yet to conclude this point. 
I have earlier suggested exposing all the tink-services via a reverse proxy. If we go via that way, we may need to only have session retained on the microservice who is exposing the tinkerbell IP, it could be nginx.

0006/README.md
Cbkhare: +1
I think, maybe will have to provide the implementation of sync logic irrespective of which db customer uses ?

Cbkhare:
@gianarb / @detiber I have updated the doc with review comments. 

I think, for some points in the proposal we would detailed layout mostly on the implementation part, which I am willing to add. Before that, I would like to know on a high-level view, does this proposal tries to solve the problem of fault-tolerance in the Tinkerbell and will it be a good feature to have?

0006/README.md
Cbkhare: done

Cbkhare:
> The title is too generic and it makes the scope of this proposal hard to discuss and understand.
> 
> my feeling is that you are confusing the work you see in docker-compose, vagrant setup with the Tinkerbell architecture itself.
> 
> A lot of the points you raised are tradeoff we had to make with docker-compose and Vagrant to provide to the community an easy way to try Tinkerbell.
> 
> Obviously it does not scale or it is not "full tolerant" but that's fine.

i have made the required changes.

0006/README.md
displague: I'm getting a similar sense from this proposal, especially with regards to the L3/L7 load balancer recommendations.  I wonder if we could start with "Recommendations for running Tinkerbell in HA", as Tinkerbell is currently provided, and use insights from that exploration to guide proposals for individual enhancements.
 
Generally, I worry that with all of the potential environments Tinkerbell could be run in, any single proposed loadbalancer (DB replication, etc) proposal may be too opinionated. We should do a better job of understanding or segmenting the landscape firs, perhaps start with a blog post or guide.

For some scenarios (dev, demo, home-lab), HA options are not important.

0006/README.md
displague: As a blog, guide, or documentation, I could see several HA avenues being described:
* part 1: deploying with kubernetes and microservices
* part 2: local loadbalancers and db replication
* part 3: cloud loadbalancers and cloud sql

Tinkerbell should work for users wanting to take any of these solutions, and perhaps many more. I agree with @gianarb that the topic of this proposal is best explored outside of the proposal system.  The shortcomings or gaps that are found in the process (why couldn't tink run in kubernetes (https://github.com/tinkerbell/tink/issues/226#issuecomment-691092198), why couldn't the database be configured to replicate, how did loadbalanced API servers miscommunicate) would lead to more concise and fitting proposals since we would better understand which users' problems we are attempting to solve.

Changes to Tinkerbell that are biased towards local-only HA solution may alienate other classes of users.

gianarb:

2. Merge it

---

### For the Author

PR #11, opened 2020-08-20
0007: sandbox and the user persona

gianarb:
Signed-off-by: Gianluca Arbezzano <gianarb92@gmail.com>

### DIFF
--- a/0007/README.md
+++ b/0007/README.md
@@ -1,6 +1,6 @@
 ---
 id: 0007
-title: PlaywithTink and the user persona
+title: Sandbox and the user persona
 status: discussion
 authors: Gianluca Arbezzano <gianarb92@gmail.com>
 ---
@@ -36,7 +36,7 @@ Non-Goals:
 ## Content
 
 I made a prototype
-[gianarb/playwithtink](https://github.com/gianab/playwithtink).
+[gianarb/sandbox](https://github.com/gianab/playwithtink).
 
 For what concerns this proposal I would like to highlight two different
 personas:
@@ -67,13 +67,13 @@ have in tink at the moment because:
 5. I think we will learn from having two different solutions for contributors
    and users those will evolve by themself as soon as they will be by their own.
 
-We can have a new repository called: `tinkerbell/playwithtink` it will contain
+We can have a new repository called: `tinkerbell/sandbox` it will contain
 a setup.sh copy/paste or equivalent script and a `docker-compose` file like the
 one we serve in `tinkerbell/tink` but with pinned docker images.
 
 We can replace or re-write setup guides similar to the one we already have, but
 they won't checkout the `tinkerbell/tink` repository, they will clone
-`tinkerbell/playwithtink` (as a release if we want extra safety) in this way the
+`tinkerbell/sandbox` (as a release if we want extra safety) in this way the
 version in use will stay the same no matter what happens in the underline
 repositories.
 
@@ -86,10 +86,10 @@ baseline for the tinkerbell.org documentation:
 * https://tinkerbell.org/docs/setup/packet-with-terraform/
 * https://tinkerbell.org/docs/setup/local-with-vagrant/
 
-`playwithtink` works in the same way but other than cloning the repo `tink`
-users will clone `playwithtink`.
+`sandbox` works in the same way but other than cloning the repo `tink`
+users will clone `sandbox`.
 
-### Playwithtink and tink/deploy are similar
+### sandbox and tink/deploy are similar
 
 Yes they are. For now. There is a bit of friction today when changing the
 content of `tink/deploy` as a contributor/maintainer because we know users that
@@ -108,6 +108,6 @@ We need a maintainer for this repository and also somebody that can actually
 make it to happen.
 
 The reponsability for the maintainers will be to bump versions up and to tag
-playwithtink releases that users can refer to when trying tinkerbell and when
+sandbox releases that users can refer to when trying tinkerbell and when
 opening a PR. From a release we should be able to identify what they run and
 replicate their issues quickly.
--- a/0007/README.md
+++ b/0007/README.md
@@ -1,0 +1,113 @@
+---
+id: 0007
+title: PlaywithTink and the user persona
+status: discussion
+authors: Gianluca Arbezzano <gianarb92@gmail.com>
+---
+
+## Summary
+
+The setup workflows and tutorial we have today in my opinion are for
+contributors. For developers who want to code for Tinkerbell.
+
+Because the process checkouts the `tinkerbell/tink` repository, and it runs the latest version of
+dependencies like: osie, tink-server, tink-worker, boots and so on. This is a
+big problem today because all those components are under development, and we
+can't guarantee that what it is in master works. Even less in integration with
+the other components. The entropy is too high.
+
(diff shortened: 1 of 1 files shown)


0007/README.md
```diff
@@ -0,0 +1,65 @@
...
+3. It is validated by the community already.
+
+We can have a new repository called: `tinkerbell/playwithtink` it will contain
+a setup.sh copy/paste or equivalent script and a `docker-compose` file like the
+one we serve in `tinkerbell/tink` but with pinned docker images.
+
+We can replace or write setup guides as similar as the one we already have, but
+they won't checkout the `tinkerbell/tink` repository, but they will `wget` the
+static files served by `tinkerbell/playwithtink` (as a release if we want extra
+safety) in this way the version in use will stay the same no matter what happens
```
gauravgahlot: The `version` here is very important. At present, the docker-compose uses `latest` tag for most of the components which, for all the right reasons, is not a good practice. 

We will have to be clear about `tinkerbell/playwithtink` in our documentation (at [tinkerbell.org](tinkerbell.org)) so that people are not confused which setup instructions to follow.

0007/README.md
```diff
@@ -0,0 +1,65 @@
...
+in the underline repositories.
+
+I see this solution as a temporary one to get something "stable" to users until
+we get a proper release lifecycle up and running.
+
+### Call for maintainers and owners
+
+We need a maintainer for this repository and also somebody that can actually
```
gauravgahlot: I will be happy to own it. :octocat:

0007/README.md
gianarb: I tried to explain as part of the RFC that we can decide if we want to replace the current setup guide with playwithtink. Is it not clear enough?! I can elaborate on it more.

gauravgahlot: Yeah, it's pretty clear. 👍🏼 
I only wanted to understand where we want to document the developer steps for tinkerbell setup. And I think we can have them with Contribution guideline under the `tink` repository.

0007/README.md
```diff
@@ -0,0 +1,106 @@
...
+Now that I have my hands dirty with the prototype I think this can be used as
+baseline for the tinkerbell.org documentation:L
+
+* https://tinkerbell.org/docs/setup/packet-with-terraform/
+* https://tinkerbell.org/docs/setup/local-with-vagrant/
+
+Those work in the same way but other than cloning the repo `tink` users will
+clone `playwithtink`.
+
+If you are concerning about playwithtink and `tink/deploy` being similar I get
```
displague: I fall under this camp :-)

0007/README.md
```diff
@@ -0,0 +1,106 @@
...
+currently have that is designed for contributors. And contributors will make
+`tink/deploy` even more efficient to better write code for Tinkerbell.
+
+### Call for maintainers and owners
+
+We need a maintainer for this repository and also somebody that can actually
+make it to happen.
+
+The reponsability for the maintainers will be to bump versions up and to tag
+playwithdocker releases that users can refer to when trying tinkerbell and when
```
displague: s/playwithdocker/playwithtink/

displague (commented):
`playwithtink` sounds like a great repository that walks a single workflow, or perhaps multiple workflows. 

This could be a series of markdown examples and assets. As a temporary, alternate, home of the deployment methods it will quickly fall out of sync without tags or immutable image shas. As an examples project, I'd be ok with being told to checkout, compile, or deploy specific commit or docker shas. The examples (readmes) would not need maintenance, they should continue to work because all of their assets and dependencies are tagged (or pseudo-tagged, as shas).

As an alternative, if we move forward with https://github.com/tinkerbell/tink/issues/228 and https://github.com/tinkerbell/tink/issues/229 we will have official homes for the terraform and vagrant deployment methods. These repositories would still need shas or tags to pin their dependencies on. Like `playontink` these projects need to be described by tinkerbell.org docs.

We've had some debate or deferal about placing 0.1.0 tags wherever we currently are in each repository. These suggestions have not been seeking a "release" tag (and a blog post about our successful build, hoorah! 🍕 ). This is just about being able to use tags as a way to reference a specific state of a software component.

If we have immutable hashes (for binary artifacts and docker images), let's use that.
If not, let's embrace tags. Let's get 0.1.0 on every project. If that number sounds too big, let's start at 0.0.1.  If we don't have CI/CD tools in place (you've been racing those in faster than we can debate their existence) then we can manually publish builds that are made locally with goreleaser or make.

With recent successes in the Terraform and Vagrant environments (tinkerbell/tink#254 and tinkerbell/tink#264), it seems that we (once again) have something working well enough to get some initial tags in place.

Don't get me wrong, I don't think 'working' is a requirement. The project is not diminished by having tags, such as "0.0.1 - nothing works", "0.0.2 - nothing works, a little better".

The longer we wait to create tags, the harder this will be. Waiting around for all of the microservices to interact perfectly in ways that we haven't even nailed down will keep us from reaching a consistent base 0.1.0 for a while.  Let's not worry about that consistent base, every project's tags will move independently and quickly.

We can stitch the various microservice tags together within `tinkerbell/tink`. `tink`'s tags can guide our major release parties and announcements. The subprojects don't have to have matching tags. Tink may be able to use hegel 0.1.0 for all of of time, while depending on osie 1.0.0 and boots 32.3.4.  I agree with what is in #5 saying asserting that these projects can have independent tags. There has been no concern raised in favor of alternate non-semver formats (like `{year}.{week}.{increment}`) which is the only thing that would give me pause about proceeding with initial tagging. 
 
Would you feel differently about all of this (tagging, release artifacts, dependency pinning) if https://github.com/tinkerbell/proposals/pull/5 were already merged? Should we get that merged first and then revisit the need for this?

gianarb:
> This could be a series of markdown examples and assets. As a temporary, alternate, home of the deployment methods it will quickly fall out of sync without tags or immutable image `shas`

We are using immutable image share here [playwithtink](https://github.com/gianarb/playwithtink)

> As an alternative, if we move forward with tinkerbell/tink#228 and tinkerbell/tink#229 we will have official homes for the terraform and vagrant deployment methods

You can see this as an attempt to have a home for both vagrant and Terraform as explained in the proposal

```
We can replace or write setup guides as similar as the one we already have, but
they won't checkout the `tinkerbell/tink` repository,
```

gianarb:
After a chat with @nathangoulding I think we can change the repo name to sandbox or playground or ...? So we will leave play with tink for what Google search uses it already

0007/README.md
```diff
@@ -0,0 +1,113 @@
...
+1. Tag a release for project, we are not ready for that and there is proposal
+   0002 under discussion about that.
+2. Getting to something that will work forever, we need something that works
+   NOW.
+3. Make the contributor experience better
+
+## Content
+
+I made a prototype
+[gianarb/sandbox](https://github.com/gianab/playwithtink).
```
gianarb: change this before merge with the actual repo tinkerbell/sandbox

displague: Also - this username has a typo (https://github.com/gianarb/playwithtink for reviewers)

displague (approved):
I pushed 88cef2f (with @gianarb's blessing and attribution as a co-author) to address some nit-pics.

This sounds like a good start to me!

PR #12, opened 2020-08-25
0008: Add proposal for common certificate handling

detiber:
## Description

Adds a proposal for common certificate handling for Tinkerbell components

## Why is this needed

Provide a unified experience between Tinkerbell components for specifying certificate configuration and to allow for configuration using both the existing certificate generation and consuming certificates that are generated with cert-manager

### DIFF
--- a/0008/README.md
+++ b/0008/README.md
@@ -1,0 +1,67 @@
+---
+id: 0008
+title: Explicit Certification For Components
+status: published
+authors: Jason DeTiberus <jdetiberus@packet.com>
+---
+
+## Summary
+
+Currently certificate configuration for Tinkerbell is handled through a
+combination of the `TINKERBELL_TLS_CERT` env variable and file based lookup.
+The file based lookup will use the `TINKERBELL_CERTS_DIR` env variable if
+defined, defaulting to `/certs/<facility>` if undefined. The certificate files
+themselves are hardcoded to `bundle.pem`, which is expected to contain both the
+CA certificate and the serving certificate, and `sserver-key.pem` for the
+private key.
+
+The current configuration methods do not offer a lot of flexibility for users,
+especially if they are looking to run Tinkerbell in Kubernetes and leverage
+automated certificate management tooling, such as
+[cert-manager](https://cert-manager.io/), for automated certificate generation
+and rotation.
+
+This proposal is intended to provide generalized guidance on how Tinkerbell
+components should expose certificate configuration to users.
+
+## Goals and not Goals
+
+Goals:
+
+1. Provide common methods of exposing certificate configuration for Tinkerbell
+services
+
+Non-Goals:
+
+1. Add tls support anywhere it is not currently enabled
+
+## Content
+
+Rather than using a combination of env variables and hard-coded defaults for
+the certificate file lookups, we can leverage explictly configured command line
+arguments for specifying the certificate path, and default certificate file
+names to use.
+
+### Expected configuration support for Tinkerbell components
+
+- Provide an optional command line argument for specifying a PEM encoded CA certificate
+  - `--ca-cert`
+  - If not provided, it will treat the provided certificate as a CA/cert bundle
+- Provide a command line argument for specifying a PEM encoded TLS certificate
+  - If TLS support is optional, then this command line argument should also be optional
+- Provide a command line argument for specifying a PEM encoded TLS private key
+  - If TLS support is optional, then this command line argument should also be optional
+- Provide an optional command line argument for specifying a PEM encoded TLS client certificate
+
+## Alternatives
+
+Keep the existing certificate handling, and implement workarounds for
+deployments/tooling that do not match up with existing expectations.
+
+Using Kubernetes and cert-manager as an example, this would require either:
+
+- Having an external process that watches the cert-manager generated secret and
+creating or updating a separate secret that is consumed by the Tinkerbell
+deployment.
+- Using an init container to create the expected files from the files that
+are mounted using the cert-manager secret


detiber:
I created a POC of using flags for configuring the tls certificates here: https://github.com/tinkerbell/tink/pull/266

gianarb:
That's great, I am not sure we should avoid the BC Break. I think we should remove the old behavior

0008/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Expected configuration support for Tinkerbell components
+
+- Provide an optional command line argument for specifying a PEM encoded CA certificate
+  - `--ca-cert`
+  - If not provided, it will treat the provided certificate as a CA/cert bundle
+- Provide a command line argument for specifying a PEM encoded TLS certificate
+  - If TLS support is optional, then this command line argument should also be optional
+- Provide a command line argument for specifying a PEM encoded TLS private key
+  - If TLS support is optional, then this command line argument should also be optional
```
mmlb, line 53: These all basically exist already. Just via env vars and automagically via use of github.com/packethost/pkg/grpc, which I thought I had example use of but doesn't look like it. https://pkg.go.dev/github.com/packethost/pkg@v0.0.0-20200807181840-a2cb6bbc41b9/grpc?tab=doc#NewServer gives some more info. IMO, tink-server should be made to use that.

detiber: @mmlb I'd have no issues standardizing on a centralized library for creating the grpc server, but if we do so, I'd prefer that it not be configured through environment variables. The environment variables create all sorts of issues when trying to build tests around anything that consumes the library.

For example, I ran into this type of issue when adding tests to boots here: https://github.com/tinkerbell/boots/pull/74/files#diff-a31be923d3cb429d14857f1b3a5a6f9f, and that just involved methods/functions within the same file, having to also be careful about env variables that might be used by transient variables for various functionality would get even messier.

detiber:
@gianarb updated to remove the backwards compatibility, I also updated https://github.com/tinkerbell/tink/pull/266 for the same change (and also updated the PR to make docker-compose work for the command line argument changes).

0008/README.md
mmlb: @detiber configuration of the lib is not solely through the env. It happens to look up in the environment for values, but supports config-setter type methods callers can use. `main` is then free to leave it to pick up env vars or grab settings from somewhere else and pass to the lib.

Its been pretty nice to have a standard GRPC setup for those apps that are using the lib, and a standard 12factor way to change when necessary.

detiber: @mmlb I'll take a closer look, though I still think it would be nice to avoid libraries interacting directly with the environment. Or at least separating out the environment handling to a separate package that provides common handling of env variables, command line flags, and possibly also tie viper (or similar) for config file support.

Not sure if @displague has additional thoughts here.

detiber: Kubernetes currently does this for flags, but I suspect with a bit of work we could do similar with env vars and config by tying into viper. 

That would allow us a few benefits:
- command line flags are more easily discoverable for users, but we could also expose env variables used through the flag description for 12-factor-based deployments
- both env variables and command line flags get a bit unweildy when there are more than a few options, so being able to use config files is nice
  - config files would also give us a nice way to reconfigure services "on-the-fly" in the case of Kubernetes-based deployments, since the config file can be mounted in through a config map, and we could either restart services on configmap update or register a file-based notifier to reload config to change config settings.

Common flag/env var/config file handling would allow us to have common handling for precedence and naming conventions across components and also avoid cases where the env variables, flags, and config settings differ between components or even potential inconsistency between the cli tools themselves and underlying libraries.

displague: I also experienced this in https://github.com/tinkerbell/tink/pull/287.

The user should be able to control the tinkerbell client from tink-worker through additional command line arguments, config files, or environment variables. When tink-worker creates a tink client (in go), the parameters should be ones that the tink worker has control over.

The current behavior `GetConnection` could still exist (and could be renamed to `client.NewClientFromEnv()`) while the desired behavior could look more like `client.NewClient(...)`. Let's say we introduced a `NewClient` function with variadic configurators:

```go
// currently client/main.go, this should be renamed client.go
package client // import "github.com/tinkerbell/tink/client

const (
  GRPCAuthorityEnv = "TINKERBELL_GRPC_AUTHORITY"
  CertificateURL = "TINKERBELL_CERT_URL"
)

type Client struct {
  GRPCAuthority, CertificateURL string
}

func FromEnv(GRPCAuthorityEnv, CertificateURL string) func (client *Client) error {
  return func(client *Client) error {
    client.GRPCAuthority = os.GetEnv(GRPCAuthorityEnv)
    client.CertificateURL = os.GetEnv(CertificateURL)
    return nil
  }
}

func GetConnection() (*grpc.ClientConn, error) {
  return NewClient(FromEnv(GRPCAuthorityEnv,  CertificateURL))
}

type ClientConfigurator func(*Client) error

func NewClient(configs ...ClientConfigurator) (*grpc.ClientConn, error) {
  client := Client{}
  for _, c := range configs {
    if err := c(client); err != nil {
      return nil, err
    }
  }

  // do the normal GetConnection things, we may need to introduce more Client attributes
  // so configurators can toggle different behaviors
  // conn := ...
  return conn, nil
}
```

displague: I've also been thinking of this from a 12-factor app perspective, but haven't called it out yet.
https://12factor.net/

gianarb:
Did we get an agreement on this? And about the bc-break? @mmlb ? I am good with it and I think we should do the bc-break

0008/README.md
mmlb: @detiber env.Get is the package wrapping getting things from the environment we could cache things there if we wanted to. I had an idea of creating a 12factor package too or something that would act as the wrapper around `env` providing a help output with a pflags like api. This was all mostly done to avoid viper since we didn't need/want config files and would rather have smaller pacakges.

But if we're going with config files now too then might as well go with viper.

0008/README.md
```diff
@@ -0,0 +1,67 @@
...
+Non-Goals:
+
+1. Add tls support anywhere it is not currently enabled
+
+## Content
+
+Rather than using a combination of env variables and hard-coded defaults for
+the certificate file lookups, we can leverage explictly configured command line
+arguments for specifying the certificate path, and default certificate file
+names to use.
```
mmlb, line 43: there's no mention of env or files to set flags, neither here nor below. It should be mentioned right?

gianarb:
As discussed via Slack we are gonna make it!! So it is ready to be merged

PR #13, opened 2020-08-26
0009: initial proposal to automate test-infra

rawkode:
## Description

Automate test infrastructure for GitHub Action repositories

## Why is this needed

If our current server fails, for whatever reason, we'd need to manually rebuild it.

### DIFF
--- a/0009/README.md
+++ b/0009/README.md
@@ -9,17 +9,61 @@ authors: David McKay <david@rawkode.com>
 
 The `tink` and `portal` repositories use GitHub Actions to run their automated tests on pull-request and push to master.
 
-A dedicated GitHub Action runner is provided for the [Tinkerbell organization](https://github.com/tinkerbell), this runner was manually created and provisioned by [Gianluca](https://github.com/gianarb); this leaves the runner vulnerable; should it suffering any problems, restoring it will be a manual process.
+A dedicated GitHub Action runner is provided for the [Tinkerbell organization](https://github.com/tinkerbell), this runner was manually created and provisioned by [Gianluca](https://github.com/gianarb). As this runner wasn't automated from the beginning, should anything happen to the server that requires it to be shutdown or migrated, it would prove timely and cumbersome to rebuild.
 
+This proposal is to use a best practice approach, Infrastructure as Code, to deliver the test infrastructure needed to support the Tinkerbell project. Doing so also provides the following benefits:
+
+- Complete transparency through open sourcing the automation to provision the runner
+- Anyone can contribute and help us maintain and improve the runner
+- Anyone can contribute runners using other cloud providers
+- Anyone can build and operate their own runner
 
 ## Goals:
 
-- Automate the creation and provisioning of the dedicated runner
+- Automate the creation and provisioning of a dedicated GitHub Actions runner
 
 ## Content: 
 
-- Create a new repository, called `tinkerbell/test-infra`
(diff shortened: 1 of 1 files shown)


0006/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+The `tink` and `portal` repositories use GitHub Actions to run their automated tests on pull-request and push to master.
+
+A dedicated GitHub Action runner is provided for the [Tinkerbell organization](https://github.com/tinkerbell), this runner was manually created and provisioned by [Gianluca](https://github.com/gianarb); this leaves the runner vulnerable; should it suffering any problems, restoring it will be a manual process.
+
+
+## Goals:
+
+- Automate the creation and provisioning of the dedicated runner
+- Convert all Tinkerbell repositories to using GitHub Actions with our dedicated runner
```
gianarb: This is a non-goal.
The runner only works for the tinkerbell/tink repository when a Vagrant CI check is required.

0006/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Future Goals
+
+We may wish to encourage all Tinkerbell repositories to utilise the dedicated runner and GitHub Actions, for consistency across the organization. Currently, all other repositories use [Drone](https://drone.io).
```
gianarb: This is a non goal imho. We should try to avoid any connection between Tinkerbell and Packet

rawkode:
@gianarb Additional goals removed

gianarb:
I would like to see a better elaboration of why we need an infrastructure as code repository (some tip):

* To outsource community needs to the community
* Because infra as code with PR and code review self-document itself
* Because automation makes the process stronger and repeatable

I was hoping for this proposal to cover how/why `test-infra` repository will look like. It is a place where we can collaborate as a community on the infrastructure required for the tink repository.

it will be nice to see reference to how you think the project should be structured (foldering), which technologies it will use and how the automation will work.

This proposal should be about building a safe workflow to collaborate on infra needs for the Tinkerbell community.
let me know if it helps

rawkode:
@gianarb I've expanded on the proposal. I am keen NOT to discuss folders and code structure, as this is an implementation detail that if we are explicit about in the proposal will constraint future refactoring, when that day comes.

0009/README.md
```diff
@@ -0,0 +1,70 @@
+---
+id: 0009
+title: Automated GitHub Action Runner
```
gianarb, line 3: Can you change this? I think we should stay a bit more general about the test-infra repo itself. The GitHub action runner is just one (probably the first) way we will use it

gauravgahlot:
@rawkode @gianarb Can we please update the PR title to include the proposal ID `0009` ?

0009/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+- Complete transparency through open sourcing the automation to provision the runner
+- Anyone can contribute and help us maintain and improve the runner
+- Anyone can contribute runners using other cloud providers
+- Anyone can build and operate their own runner
+
+## Goals:
+
+- Automate the creation and provisioning of a dedicated GitHub Actions runner
+
```
displague, line 24: The following motivator was mentioned in the community meeting https://youtu.be/QzK_aJMYqpk (22m:40s @gianarb)
`- Host bots and other custom resources (such as triage-party)`

gianarb:

2. Merge it

---

### For the Author

PR #14, opened 2020-09-01
0010: on_timeout and on_failure as root elements of a workflow

gauravgahlot:
## Description
Add `on_timeout` and `on_failure` actions as root elements of a workflow.

## Why is this needed
To add better support for what a user can do when a workflow fails or timeout.

### DIFF
--- a/0010/README.md
+++ b/0010/README.md
@@ -1,7 +1,7 @@
 ---
 id: 0010
 title: Workflow behavior for on_timeout and on_failure
-status: discussion 
+status: abandoned 
 authors: Gaurav Gahlot <gauravgahlot0107@gmail.com>
 ---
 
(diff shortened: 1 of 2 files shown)


gianarb:
I have a question if we create a good event-driven mechanism that a client can use and subscribe to, it will get notified by a timeout or error and it will act by consequence OUTSIDE from Tinkerbell. I am worried about this will turn to the complexity that we can't leave to somebody else.

gauravgahlot:
> I have a question if we create a good event-driven mechanism that a client can use and subscribe to, it will get notified by a timeout or error and it will act by consequence OUTSIDE from Tinkerbell. I am worried about this will turn to the complexity that we can't leave to somebody else.

Based on the discussion in the last Tinkerbell Contributors Meeting, we would like to take this opportunity and make tinkerbell more event-driven than what it is today. 

The proposal has been marked as `abandoned` as it is not relevant at the moment. However, we may revisit and redefine the proposal based on the new event-driven architecture.

PR #15, opened 2020-09-02
0011: how to communicate bc-break

gianarb:


gauravgahlot:
I don't have any prior experience with OSS release. 

**Questions:**
- _Is it a practice around OSS release to add bc breaks in release notes?_
- _If not, does adding bc breaks to release notes make sense?_

gianarb:
Yes, it is a good practice. It is not that common because a big project does not have bc breaks after v1.0.0 but when a project starts it is important to create an effective feedback loop with early adopters.

* https://github.com/kubernetes-sigs/cluster-api/releases

gauravgahlot:
LGTM 👍🏼 🚀

0011/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+Every PR that introduce a BC break will get labeled with a `bc-break` label. In
+this way we will be able to build a workflow that programmatically will extract
+the BC-Break for every component.
+
+Every PR has to contain as part of the description a section titled:
+"how to migrate" teaching the user about how it should pass from a previous
+version to a new one when possible. If not possible because the section will
+explain why we remove the entire functionality and will offer an alternative if
+it exists.
```
mmlb, line 54: I think even better would be to have a code snippet be a soft-requirement, similar to how `go fix` used to exist to handle this before Go1. I don't think it needs to be a specific tool, or expected to always be 100% correct dealing with odd OSes (non-gnu sed for example). It would a serve the purpose of "how to mitigate" by showing with code that can either be run or "ported".  Nothing too onerous either, sed, awk, bash, python (maybe perl), even go code itself maybe.

Maybe something like https://en.wikipedia.org/wiki/Coccinelle_(software) or similar would be a good option. I feel like there's something along the lines of coccinelle but easier/friendlier language maybe, but I can't seem to find anything.

gianarb: Yep! That's nice. I don't want to go too deep on "How the migrate section should be written" mainly because I can't really guess all the possible bc-break. Maybe the bc is an HTTP API field, maybe a code one, maybe a deprecated feature, so it won't be possible to migrate.

Who knows! It looks like a deep rabbit hole to follow right now. But I think reviewers and lesson learned plays a good lesson here. WE have to be sure that the migrate chapter is well written, what ever it means

mmlb: Yeah those are good points. Somethings can't be easily specified in code. But I think having a call out to add a script or similar when it *is* possible would make things easier. For example https://github.com/tinkerbell/tink/pull/280 can have a sed script adding a default timeout.

0011/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## how to migrate
+
+The "how to migrate" section has to be placed in the PR description. Here a few
+things you should include:
+
+1. You have to be clear about the required steps.
+2. You can write a bash scripts if necessary that can be used to perform
+   the migration.
+3. This chapter is the documentation users will relay on when performing an
```
mmlb:
```suggestion
3. This chapter is the documentation users will rely on when performing an
```

0011/README.md
```diff
@@ -0,0 +1,87 @@
+---
+id: 0011
+title: how to announce bc break for Tinkerbell
```
mmlb: bc isn't ever defined. I suggest you do the full wording in the title and an explicit call out that "in the rest of this document bc stands for backwards compatibility"

0011/README.md
```diff
@@ -0,0 +1,87 @@
...
+status: discussion
+authors: Gianluca Arbezzano <gianarb92@gmail.com>
+---
+
+## Summary
+
+At this stage, where the various services are not released and do not have
+semver we can accept BC break.
+
+We need to have a consolidate way to communicate them.
```
mmlb:
```suggestion
We need to have a consolidated way to communicate them.
```

0011/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Goals and no-Goals
+
+Goal:
+
+* Clarify the way we can communicate bc break across services and in Tinkerbell.
```
mmlb:
```suggestion
* Clarify the way we can communicate bc break across services in Tinkerbell.
```

0011/README.md
```diff
@@ -0,0 +1,87 @@
...
+* A solution with low overhead that has to work NOW.
+
+No-Goal:
+
+* Explain how SemVer manage Bc-Break or how we will communicate bc-break when
```
mmlb:
```suggestion
* Explain how SemVer manages bc-break or how we will communicate bc-break when
```

0011/README.md
```diff
@@ -0,0 +1,87 @@
...
+explain why we remove the entire functionality and will offer an alternative if
+it exists.
+
+## how to migrate
+
+The "how to migrate" section has to be placed in the PR description. Here a few
+things you should include:
+
+1. You have to be clear about the required steps.
+2. You can write a bash scripts if necessary that can be used to perform
```
mmlb:
```suggestion
2. You can write a bash script if necessary that can be used to perform
```

0011/README.md
```diff
@@ -0,0 +1,87 @@
...
+   the migration.
+3. This chapter is the documentation users will relay on when performing an
+   update. You should explain how to validate if the migration went right.
+4. If a bash script is not enough feel free to implement a new binary, even
+   better if it has a `--dry-run` approach that can be used to validate the
+   action that will be performed without having to do them straight away.
+5. We know that updates are crucial, and critical, sometime is better to have
```
mmlb:
```suggestion
5. We know that updates are crucial, and critical, sometime it is better to have
```

0011/README.md
```diff
@@ -0,0 +1,87 @@
...
+this way we will be able to build a workflow that programmatically will extract
+the BC-Break for every component.
+
+Every PR has to contain as part of the description a section titled:
+"how to migrate" teaching the user about how it should pass from a previous
+version to a new one when possible. If not possible because the section will
+explain why we remove the entire functionality and will offer an alternative if
+it exists.
+
+## how to migrate
```
mmlb:
```suggestion
## How To Migrate
```

0011/README.md
```diff
@@ -0,0 +1,87 @@
...
+4. If a bash script is not enough feel free to implement a new binary, even
+   better if it has a `--dry-run` approach that can be used to validate the
+   action that will be performed without having to do them straight away.
+5. We know that updates are crucial, and critical, sometime is better to have
+   the best written format of what it is required vs a too magic bash script
+6. Something like
+   [coccinelle](https://en.wikipedia.org/wiki/Coccinelle_(software) or [`go
+   fix`](https://golang.org/cmd/fix/) should be evaluated as well.
+
+## share and aggregate bc breaks
```
mmlb:
```suggestion
## Share And Aggregate BC Breaks
```

mmlb (changes requested):

0011/README.md
```diff
@@ -0,0 +1,92 @@
+---
+id: 0011
+title: how to announce a braking change for Tinkerbell
```
mmlb:
```suggestion
title: how to announce a breaking change for Tinkerbell
```

mmlb (commented):
last one, and then :+1:

PR #17, opened 2020-09-14
0012: Event driven tinkerbell

gauravgahlot:


0012/README.md
```diff
@@ -0,0 +1,186 @@
...
+id: 0012
+title: Event driven design for Tinkerbell
+status: ideation
+authors: Gaurav Gahlot <gauravgahlot0107@gmail.com>, Gianluca Arbezzano <gianarb92@gmail.com>
+---
+
+## Summary
+
+In the current model, the whole tinkerbell stack works on a _request-response_ model.
+Instead of waiting for explicit requests or responses, we want the components to watch for certain events and act accordingly.
```
mmlb: This line doesn't really match the non-goal of "Move Tinkerbell in the event-driven design direction"

0012/README.md
```diff
@@ -0,0 +1,186 @@
...
+No-Goal:
+
+- Disrupt the current working environment.
+- Move Tinkerbell in the event-driven design direction.
+- Update workers to receive actions via event driven model.
+
+## Content
+
+Not everything can be a Tink responsibility.
+Events are a scalable way to build an extendible system.
```
mmlb:
```suggestion
Events are a scalable way to build an extensible system.
```

0012/README.md
```diff
@@ -0,0 +1,186 @@
...
+}
+```
+
+So when we create a `HardwareClient`, for instance, it would return a watcher for the `Hardware` resource type.
+All the clients like `HardwareClient` will be using `EventClient` under the hood.
+
+There is no logic to keep track of which events are fired or sent to a consumer.
+Every consumer when it connects to a stream of events will specify how old the events returned should be (by default 5m).
+
+The consumer will have to implement the business logic in an idempotent way.
```
nathangoulding: I don't think tink should count on the consumer implementing things in a certain way. There may be a robust queue / broker / exchange / routing service downstream of the tink-server, but tink should not implement any of this logic itself. It should simply stream events to any of its watchers.

0012/README.md
```diff
@@ -0,0 +1,186 @@
...
+Forcing the consumer to implement the logic in a way that can be run repeatedly simplifies the logic in Tink but it also enforce reliability in the client side.
+If a client fails half way the client gets the same events again and repeat the logic.
```
nathangoulding, line 222: Just to reiterate, I don't think replaying or other queueing logic should be built into tink-server.

0012/README.md
gianarb: That's what the message says. We can rephrase it but, you are right. Tink only streams events. The consumer has to figure out what to with them and to deal with eventually getting the same events twice (if it reconnects to the stream)

0012/README.md
gianarb: It is not a queuing system. Events are stored in Postgres and by default when a stream start, it replays events from 5 minutes ago. The consumer can set the time to 0 but probably it will lose the events that happened in between the reconnection. You have to know that I think there is too much code in the world already, and I like minimalism.

0012/README.md
```diff
@@ -0,0 +1,186 @@
...
+title: Event driven design for Tinkerbell
+status: ideation
+authors: Gaurav Gahlot <gauravgahlot0107@gmail.com>, Gianluca Arbezzano <gianarb92@gmail.com>
+---
+
+## Summary
+
+In the current model, the whole tinkerbell stack works on a _request-response_ model.
+Instead of waiting for explicit requests or responses, we want the components to watch for certain events and act accordingly.
+This introduces extensibility and allows users to plug-and- play with tink.
```
Cbkhare: Would like to understand more about what events are we talking about and how they are related to `plug-and-play`.

also, there is a typo.

0012/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+- Disrupt the current working environment.
+- Move Tinkerbell in the event-driven design direction.
+- Update workers to receive actions via event driven model.
+
+## Content
+
+Not everything can be a Tink responsibility.
+Events are a scalable way to build an extendible system.
+This allows different components to tap-in to the event streams and leverage the extension points.
```
Cbkhare, line 36: how this `event stream` would be different from `pub-sub` message model.

0012/README.md
```diff
@@ -0,0 +1,186 @@
...
+For example, [tinkerbell/portal](https://github.com/tinkerbell/portal/) can watch for workflow events and present them on the UI.
+Events are good for troubleshooting purpose because they help to build context.
+Feature requests we received that can be implemented using events are: phone home, ontimeout and so on
```
Cbkhare: I would like to have an example here. In a lifecycle of an event, how will it interact with the tink components or vice-versa?

0012/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Current Model
+
+In the current model, the tink-server is loaded with tons of responsibilities.
+Responsibilities like - managing workflow state, handing overs actions to workers, logging the events, and others.
+We would like to delegate these responsibilities to respective smaller services/components.
```
Cbkhare: need some more explanation on components. Where will they live in tink architecture?
Secondly, how will these components send/receive the events from the grpc, will it be redirected by tink-server ?

0012/README.md
```diff
@@ -0,0 +1,186 @@
...
+All the clients like `HardwareClient` will be using `EventClient` under the hood.
+
+There is no logic to keep track of which events are fired or sent to a consumer.
+Every consumer when it connects to a stream of events will specify how old the events returned should be (by default 5m).
+
+The consumer will have to implement the business logic in an idempotent way.
+Forcing the consumer to implement the logic in a way that can be run repeatedly simplifies the logic in Tink but it also enforce reliability in the client side.
+If a client fails half way the client gets the same events again and repeat the logic.
+
+The `Watch` gRPC function needs to support filtering by a `ResourceType` and/or a `ResourceID` and/or an `EventType`.
```
Cbkhare, line 234: Is this `Watch` function only going to make queries from tink-server and expects consumers to revert on the basis of the type of components?

0012/README.md
```diff
@@ -0,0 +1,186 @@
...
+- `data`: the primary event payload, represented as `interface{}`.
+For a workflow created event, the payload (data) can be the complete workflow structure.
+For an on-timeout event the data can be an action that needs to be executed next.
+
+### EventClient
+
+A new `EventClient` should be developed with the primitive required for the events at least `Watch`.
+At the beginning, only a `Watch` function is required but we think a natural evolution will be to serve a function that can be used from other components to fire events.
+Or maybe we can do a `Watch` and `Create` straight away.
+All the resource clients will have a `Watch` function that will stream events for that particular resource.
```
Cbkhare, line 173: After streaming the events what this `Watch` function (or the method which triggered this `Watch`), intends to do?

0012/README.md
Cbkhare: wouldn't this end up in flooding all the events onto all the consumers? or if you are planning to redirect the event to the intended consumer, maybe with MQ or over the network?

0012/README.md
Cbkhare: Which component of tink will manage the trigger of streams?

0012/README.md
gianarb: Pub-Sub is fire and forget. This events are persisted and can be replayed (until a TTL is reached). Anyway, pub-sub is a pattern, stream is a communication technique I presume. So I can't really answer

0012/README.md
gianarb: Those 3 are examples. We can detail them a bit more. I can't find public documentation around the phone call feature @nathangoulding . But @Cbkhare I presume you know what on-timeout means.

I can detail a bit more how the `on-timeout` can be implemented:

1. You call: (this is pseudocode)

```golang
informer := client.WorkflowClient.Watch(&request.WatchRequest{
    EventType: "WORKFLOW_TIMEOUT"
}, func(e Event) {
   // Do your best. You can notify via Slack. Or start a different workflow.
})
informer.Run(ctx)

File: 0012/README.md (hunk 8 above)
Comment (gianarb) [reply 0012/README.md]: > need some more explanation on components. Where will they live in tink architecture?

it is not part of this proposal. The separation of concerns won't change here. We are just creating a mechanism that enables third parties to implement their own logic outside of Tinkerbell.

There is no redirection, as you can see in the system graph, events are stored in Postgres and there is a couple of new gRPC request served by the grpc server.

File: 0012/README.md (hunk 5 above)
Comment (gianarb) [reply 0012/README.md]: There is a list of events, not all of them but some. Right now events are CRUD around the tink resources and workflow state

File: 0012/README.md (hunk 4 above)
Comment (gianarb) [reply 0012/README.md:222]: tink-server because that's where the logic happens :) Tink Server is the component that today receives requests for resource manipulation (create, delete, update hardware or template or workflow). It also manages the workflows state. That's where the events will come from.

**Comment by gianarb** [comment]
@Cbkhare I can't replay for those inline for some reason.

> Which component of tink will manage the trigger of streams?

The stream is a grpc stream. The grpc server is currently served by tink-server. I think the answer is tink-server.

File: 0012/README.md (hunk 10 above)
Comment (gianarb) [reply 0012/README.md:173]: It returns something that waits and runs your code when an event gets fired.

File: 0012/README.md (hunk 9 above)
Comment (gianarb) [reply 0012/README.md:234]: `only`? 👍

File: 0012/README.md (hunk 3 above)
Comment (gianarb) [reply 0012/README.md]: This system is not event heavy. We are speaking about really a low number of events. Like a couple of thousands per hour. The TTL is the safeguard. Are you managing too many events?! Make the TTL shorter. The queue if required can be implemented outside of Tink. In this way, we do not have to maintain a list of implementations for different queuing systems. People have opinions and there are too many rabbitmq, redis, kafka, zeromq out there 🍡  There are other backpressure mechanisms that can be implemented in the future if required to speed up how events are written or streamed.

File: 0012/README.md [hunk 11]
```diff
@@ -0,0 +1,215 @@
...
+
+## Implementation Details
+
+### PostgreSQL Notifications
+
+Tinkerbell uses PostgreSQL as the data store.
+Postgres provides [notifications](https://www.postgresql.org/docs/10/sql-notify.html) which can be used to watch the changes in a table as they occur.
+Here changes refer to Postgres events - INSERT, UPDATE, and DELETE.
+
+NOTIFY provides a simple interprocess communication mechanism for a collection of processes accessing the same PostgreSQL database. A payload string can be sent along with the notification, and higher-level mechanisms for passing structured data can be built by using tables in the database to pass additional data from notifier to listener(s).
```
mmlb:
```suggestion
NOTIFY provides a simple interprocess communication mechanism for a collection of processes accessing the same PostgreSQL database.
A payload string can be sent along with the notification, and higher-level mechanisms for passing structured data can be built by using tables in the database to pass additional data from notifier to listener(s).
```

0012/listener.go
```diff
@@ -0,0 +1,65 @@
+package listener
+
+// import (
```
mmlb: why is this all commented out?

0012/README.md
```diff
@@ -0,0 +1,215 @@
...
+    ResourceID // the resource ID (ResourceType required when this is set)
+    ResourceType // workflow, template, action, hardware
+}
+```
+
+So when we create a `HardwareClient`, for instance, it would return a watcher for the `Hardware` resource type.
+All the clients like `HardwareClient` will be using `EventClient` under the hood.
+
+There is no logic to keep track of which events are fired or sent to a consumer.
+Every consumer when it connects to a stream of events will specify how old the events returned should be (by default 5m).
```
mmlb, line 199: The stream isn't being persisted in any of the examples here. That should be addressed in this proposal no?

0012/README.md
```diff
@@ -0,0 +1,215 @@
...
+The new model is inspired from Kubernetes.
+The resource clients should have a `Watch` function that will stream events for that particular resource.
+
+```
+Watch(ctx context.Context, in EventWatchRequest)
+```
+
+## Implementation Details
+
+### PostgreSQL Notifications
```
mmlb, line 110: This doesn't really make sense at this point in time. Since PG Notify isn't persisted then there's no replay ability. Which means that needs to be handled at a higher layer. So what exactly are we gaining from using NOTIFY/LISTEN? What we'd really need if we wanted to do it in the DB is something more like [auditing](https://wiki.postgresql.org/wiki/Audit_trigger_91plus) would be needed.

mmlb (changes requested):

0012/README.md
gianarb: Yes, I think this won't be the real implementation, yesterday evening I was thinking about the possibility to implement it differently. 

We will persist events directly in a DB table from tink-server. Where they happen. With an insert. And I think we will use listeners only on the event table to get the new line and stream them out.

In the meantime we will look at the auditing feature as well. My direction at this stage is to be 70% focused on a good grpc layer, and the rest is "make it working". When it works we can figure out how to make it better. But probably the v1 will be as I wrote, at some point in the code a new event will be stored in the event table.

0012/README.md
Cbkhare: i mean, are other components going to leverage this `Watch` for some other task as well ? Maybe after getting response from consumer it can push some data to logging or monitoring stack. ?

0012/README.md
Cbkhare: okay, to understand better I was trying to create an analogy between `event stream` as a part of Event-driven pattern and pub-sub model and also, its difference with later. 

Reading this file and comment in the link below, i think what you are trying to implement is `Event sourcing` https://martinfowler.com/eaaDev/EventSourcing.html pattern and not `Event-driven` pattern? 

https://github.com/tinkerbell/proposals/pull/17/files/b71317a6ddca4227ec1b0394d99afcfcb7cd3083#r488554642

0012/README.md
Cbkhare: I assume this would be an async implementation instead of polling. I would request a brief idea on how this `wait and run on event` model would be implemented, ex how are you planning to keep a thread/process open/sleep/aync until an event is received or would it be similar to RMI or leverage some go library(some guess)?

0012/listener.go
gauravgahlot: I have uncommented the code.

0012/README.md
gauravgahlot: I have updated the proposal with details about the events table, where all the generated events will be stored.

0012/README.md
Cbkhare: okay, I would request some details about this `delegate these responsibilities to respective smaller services/components` , may be with an example.

0012/README.md
gauravgahlot: oops 👀

0012/README.md
```diff
@@ -0,0 +1,233 @@
...
+ event_type    | integer                  |           | not null |
+ created_at    | timestamp with time zone |           |          |
+ data          | jsonb                    |           |          |
+Indexes:
+    "events_pkey" PRIMARY KEY, btree (id)
+```
+
+The tink-server will only be responsible for generating the events.
+It will not contain any business logic that needs to be executed as an event occurs.
+Instead it's the consumer who will have to implement the business logic in an idempotent way.
```
Cbkhare, line 220: how are we handling the scenario where an event was triggered but consumer didn't received it ?
In that case, neither consumer is aware if event was triggered nor tink is aware if event was received ?
Also, a possible of re-delivery.

Cbkhare (changes requested):

0012/README.md
```diff
@@ -0,0 +1,233 @@
+---
```
Cbkhare, line 1: Some of my comments are marked as resolved. I would request some reply on them.

Cbkhare (changes requested):

0012/README.md
gauravgahlot: Adding a delivery/retry logic will be an overkill at this point. Every consumer when it connects to a stream of events will specify how old the events returned should be (by default 5m). That way they will receive the events missed during the connect breakdown.

0012/README.md
gauravgahlot: Watch is going to be as simple as an infinite `for` loop running in a goroutine. All the events as they occur will be stored in the events table and will be streamed over gRPC and it will be up to the consumer how they want to consume the stream.

0012/README.md
gauravgahlot: As the proposal speaks, and you rightly said, the components are free to `Watch` for the resource changes and do certain operations. Logging, and monitoring could also be implemented on similar lines.

0012/README.md
gauravgahlot: That was unintentional actually. I was marking resolved those we have responded to and accidentally marked your new comments as well. 😄

0012/README.md
Cbkhare: I don't agree it would be an overkill. 

Consumer is unaware if an event was triggered. Tink will assume it should have been received by the consumer. What if consumer never asks or when it asks, then some events are left out for not being the part of the `time-period`. Consumer will endup being inconsistent and can't be blamed since it is unaware if an event was triggered.

Cbkhare (changes requested):

0012/README.md
Cbkhare: np, i understand.

0012/README.md
gauravgahlot: @gianarb and I are working to add more examples and implementation details so that it's clear what we intend to do here.

0012/README.md
gianarb: > how are we handling the scenario where an event was triggered but consumer didn't received it ?

We do not have a catalog of consumers. Consumers do not register themself (consumers are normal clients like the tink-cli) if a consumer is not attached to a stream or if it didn't ask for old enough events for the tink-server that consumer does not exist. When a consumer gets implemented it knows that tink-server stores events for some amount of time, it is the responsibility of who implements the consumer to figure out how to be reliable. 

The tink-server responsibility is to store and stream events reliably and fastly to whoever asks for them

gianarb:
@Cbkhare I deleted a comment by mistake but I think you posted this twice: https://github.com/tinkerbell/proposals/pull/17#discussion_r489275832

0012/README.md
Cbkhare: I think my comment was deleted. Let me add it again. 

Event is streamed to consumer when it occurs and also, consumer is listening onto the stream and is unaware about the occurrence of an event until it receives it.

> The tink-server responsibility is to store and stream events reliably and fastly to whoever asks for them

How are we ensuring the streaming events reliably, what happens when the event is lost during the stream, may be due to network issue or service restart? Just like with a REST call we are aware about the status of the execution with status-code and resend it or take corrective actions. 

Maybe this should help https://sourcegraph.com/github.com/grpc/grpc-go/-/blob/stream.go#L1331.

0012/README.md
mmlb: ok makes sense. The nice thing about the trigger based approach is that the DB takes care of populating the events table itself and we don't have to worry about sprinkling code in everywhere we want an event. I'd really consider doing it that way first actually.

mmlb: @Cbkhare I think this should answer your question. The events need to be persisted into a table. NOTIFYs are from there, new watchers would SELECT to go back in time and then LISTEN for new updates. Some care would be needed to make sure we don't race and drop some events between the two.

0012/README.md
mmlb: I think this is answered in https://github.com/tinkerbell/proposals/pull/17/#discussion_r488916179

mmlb:
A few of the comments are about the future direction, and maybe premature at this point. A lot of confusion/questions are coming up on it but its out of scope I think. Suggest the proposal is re-worded a bit to just mention future possibilities as lightly as possible.

gianarb (approved):

0012/README.md
```diff
@@ -0,0 +1,260 @@
...
+    ResourceID // the resource ID (ResourceType required when this is set)
+    ResourceType // workflow, template, action, hardware
+}
+```
+
+So when we create a `HardwareClient`, for instance, it would return a watcher for the `Hardware` resource type.
+All the clients like `HardwareClient` will be using `EventClient` under the hood.
+
+There is no logic to keep track of which events are fired or sent to a consumer.
+Every consumer when it connects to a stream of events will specify how old the events returned should be (by default 5m).
```
mmlb, line 199: I wonder if this is going to be a lot of work on the clients that would make them more complex.

Cbkhare (approved):

0012/README.md
gianarb: Can you be a bit more precise? It is just a parameter, if they don't want to get old events they can ask for `0s`. I think by default we should avoid the possibility for a client to lose events during reconnection (for example during a consumer's deploy)

gianarb:
I am gonna merge this as it is. It is open for future discussions if something has to change or if it will diverge from the actual implementation. But for now, I think it is good enough

PR #18, opened 2020-09-23
0013: Propose Self Contained Workflows

mmlb:
## Description

Proposes the concept of fully self-contained workflows. Its used as the reverse-dependency of [0014](https://github.com/tinkerbell/proposals/pull/19) and [0015](https://github.com/tinkerbell/proposals/pull/20). 

## Why is this needed

Will allow for better managed workflows that can fully lock down versions of all of its necessary files. Also allows to remove OS/machine specific code from Boots.

PR #19, opened 2020-09-23
Propose Workflow Services

mmlb:
## Description

Sets up a mechanism to run containers in the background, there by providing a service the task actions will be able to use.

## Why is this needed

Allows for ripping out code from Boots and enables self-contained workflows ([0013](https://github.com/tinkerbell/proposals/pull/18)).

### DIFF
--- a/0014/README.md
+++ b/0014/README.md
@@ -1,0 +1,100 @@
+---
+id: 0014
+title: Workflow Services
+status: published
+authors: Manny <manny@packet.com>
+---
+
+## Summary
+
+A Workflow Service is part of a Workflow, just like Tasks.
+It is a container that starts before any Task and runs in the background.
+Services are *not* required to continue running for the entire duration of the workflow.
+Once the Workflow is Done (Finished/Errored/Timedout...) it will be killed/cleaned up.
+
+Support for running workflow actions in the background.
+This is to support [0013] in conjuction with [0015].
+
+## Goals and not Goals
+
+Run containers in detached/background so they may provides "services" to the further workflow actions/tasks.
+
+## Content
+
+By having containers running in the background we remove "global" code from boots and have it local to the workflow instead.
+The same benefits could be had for serving of the installer/image files.
+This is the initial/immediate use case, I would not be surprised to see others come up.
+
+I initially thought about making this an attribute of the action.
+That seemed a little bit too loose and easy to make bad choices with so changed it to a separate kind of "job" for the task.
+@nathangoulding suggested a top-level attribute instead as it most likely makes sense to run this for the full life time of a workflow.
+This makes sense to me.
+It seems like the service should not out live its parent, so we would want to terminate services at the end of a task if we were still under a task.
+
+#### Lifecycle
+
+The services will be started before any Tasks.
+All services will be required to be `Up` and `healthy` (if HEALTHCHECK is setup in the image) before any tasks are started.
+Services are *not* required to continue running for the entire duration of the workflow.
+Attempts to restart the service may occur if a service exits with an error.
+A service being down will only affect the Workflow state if a Task/Action itself fails due to the missing service.
+Once the Workflow is no longer Active (Errored/Timedout/Finished) all services will be terminated.
+
+The Tink Event Bus will be needed in order to make use of Task/Action results.
+
+```yaml
+version: "0.1"
+name: ubuntu_provisioning
+global_timeout: 6000
+services:
+  - name: serve boot files
+    worker: {{provisioner1}}
+    image: osie-boot-files-server:v1.0.42
+    labels:
+      - "traefik.http.routers.myrouter.rule=Host(`{{workflowid}}.workflows.tinkerbell.local`)"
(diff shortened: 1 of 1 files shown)


mmlb:
Rendered: https://github.com/mmlb/proposals/blob/0014/0014/README.md

gianarb:
This is reasonable and for some reason, the syntax brings me to LinuxKit for some reason

https://github.com/linuxkit/linuxkit/blob/master/examples/influxdb-os.yml#L13

Anyway, I like this proposal but I think we should do a better exercise at naming things. We can call this feature: "Service" and describe it clearly as: "A service is part of a workflow, just like actions. It is a container that runs in the background for the all workflow lifecycle".

Something we should write down and define
1. how will it affect the workflow execution if it errors or times out? I think it should work the same as an action.
2. will it be killed when a workflow reaches its global timeout but obviously, it won't hold(1) a workflow.

I think we should not hold this proposal too much, I would like to see it published even if the other 2 are not ready, it is self-contained and reasonable by its own.

(1): Only actions and tasks count in terms of execution time, services just runs all the time.

mmlb:
> This is reasonable and for some reason, the syntax brings me to LinuxKit for some reason
> 
> [https://github.com/linuxkit/linuxkit/blob/master/examples/influxdb-os.yml#L13](https://github.com/linuxkit/linuxkit/blob/master/examples/influxdb-os.yml?rgh-link-date=2020-09-24T14%3A34%3A57Z#L13)

I'm more familiar with https://0-8-0.docs.drone.io/services/ which I think has a lot of similarities with our workflows, the need/existence there and the use we have points to a fundamental basic requirement for this.

> Anyway, I like this proposal but I think we should do a better exercise at naming things. We can call this feature: "Service" and describe it clearly as: "A service is part of a workflow, just like actions. It is a container that runs in the background for the all workflow lifecycle".
> 

Part of the reason I like proposals :D. I'm pretty bad at naming things (*cough* cacher *cough*) and opening that up before code exists makes it easier to ask/provide that kind of feedback.

This wording sgtm, will add.

> Something we should write down and define
> 
>     1. how will it affect the workflow execution if it errors or times out? I think it should work the same as an action.

I don't really have a strong opinion. But, consider what if the use of the service is done when it encounters an error? If its not ever going to be used again I don't think it matters. If it is used again then the task will most likely fail because of it anyway.

> 
>     2. will it be killed when a workflow reaches its global timeout but obviously, it won't hold(1) a workflow.

Yes, services are tied to a workflow lifetime, so when the workflow is done (success/fail/timeout) the services are killed.

> 
> I think we should not hold this proposal too much, I would like to see it published even if the other 2 are not ready, it is self-contained and reasonable by its own.

Yep, part of the reason for not doing a jumbo-proposal.:+1:

> 
> (1): Only actions and tasks count in terms of execution time, services just runs all the time.

Yeah I think this makes sense, I expect all services to be up and healthy (I'll add this in) before starting the tasks. I suspect service startup time won't be too long.

0014/README.md
```diff
@@ -26,8 +31,15 @@ That seemed a little bit too loose and easy to make bad choices with so changed
...
 It seems like the service should not out live its parent, so we would want to terminate services at the end of a task if we were still under a task.
 
-The services would be running before any tasks are started, and will be killed at the end of the workflow.
-The service would thus need to be a client of the tinkerbell stream if it wanted to react to any state changes/events.
+#### Lifecycle
+
+The services will be started before any Tasks.
+All services will be required to be `Up` and `healthy` (if HEALTHCHECK is setup in the image) before any tasks are started.
+Services must be `Up` for the duration of the Workflow.
+Just like Actions, if a Service exits prematurely the Workflow will be considered Errored.
```
mmlb: @gianarb here we're going with what you said. I still kind of think maybe its ok to let the workflow continue on if a service that is no longer needed errors.

I intentionally worded it to have the services up the whole duration of the workflow. We could be ok with letting the service exit gracefully though too.

gianarb: I think you are right; I didn't have an opinion about that; I was just asking. 
The point you raised has a lot of sense; if an action requires a service that is unhealthy, the action will fail the workflow will follow. If a service is down, but nobody needs it, why should we mark the workflow as a failure? I think it has sense :) I was far from having a solid opinion about that.

mmlb: Cool lets do that then.

gianarb:
I think you have to sign your commit @mmlb

mmlb:
hmm I thought I did

gianarb (approved):
I think we should change the status of this to `published` and merge it!

mmlb:
Doing it!

mmlb:
oh wait @gianarb I didn't update the bit about "exit gracefully" or "letting the task/action" discover a dead service and fail. Going to do that and let you do a final review on it.

mmlb:
@gianarb I've pushed new wording making services not be required to be up. I think we might just end up running with `--restart=on-failure`, but that should stay an implementation detail I think.

PR #20, opened 2020-09-23
0015: Propose HTTP Routing for Services

mmlb:
## Description

Defines a way to setup HTTP routing for services.

## Why is this needed

Makes background workers/services ([0014](https://github.com/tinkerbell/proposals/pull/19)) actually useful :D.


mmlb:
Rendered: https://github.com/mmlb/proposals/blob/0015/0015/README.md

0015/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Summary
+
+HTTP routing for services started by tink-worker running on "provisioner" machines.
+
+## Goals and not Goals
+
+#### Goal
+
+A way for provisonee/target machines to make HTTP(S) requests to a service that is defined as part of its workflow.
```
gianarb: In Tinkerbell, we call them `hardware`. let's reduce the number of terms we use when possible. If I well understood the proposal this code will probably live in `tink-worker`. Probably worth mentioning in order to increase context.

gianarb:
I think a diagram will help to clarify the use case. I think with the information and knowledge I have I can't really understand this proposal, can you write a `use case for dummy` as part of the proposal?

I am wondering why HTTP/TCP as a way to exchange those files vs local file systems. Why we do not create a mechanism that copies file from inside a container to the host and/or with volumes?

How an action will know the URL? With the proposed link it has to know not only the service it interact with (reasonable) but also the `workflowid` (less reasonable)?

If we make better use of the Docker network and the DNS we can probably get to something a bit more dynamic like: one network per workflow and internally actions and services (containers) can reach out to each other via name (more or less like docker-compose).

0015/README.md
mmlb: Well yes, but I want to separate different kinds of hardware/workers. Is hardware always meant to be the machine that is being installed/most-of-the-workflow-is-actually-for?

mmlb:
> I think a diagram will help to clarify the use case. I think with the information and knowledge I have I can't really understand this proposal, can you write a `use case for dummy` as part of the proposal?
> 

Can do.

> I am wondering why HTTP/TCP as a way to exchange those files vs local file systems. Why we do not create a mechanism that copies file from inside a container to the host and/or with volumes?
> 

This already exists I think, in statedir/ephemeral_data. This proposal is really for when 2 machines are working together on a workflow (in the yaml `{{provisioner1}}` and `{{device_1}}`). Though should still be usable if only on 1 machine so that a service doesn't need to be ported from native http -> localfile/ephemeral_data. 

> How an action will know the URL? With the proposed link it has to know not only the service it interact with (reasonable) but also the `workflowid` (less reasonable)?

Does the action not know its own workflowid? That seems like a useful bit of metadata to always know (even if just for logging/context purposes). Knowing the action is out of scope for this proposal, but we have lots of places where we embed/inject a url (metadata, boots filename)...

> 
> If we make better use of the Docker network and the DNS we can probably get to something a bit more dynamic like: one network per workflow and internally actions and services (containers) can reach out to each other via name (more or less like docker-compose).

Yep this would be fine for the local machine traffic only, but I'm looking for this more for multi-machine workflows.

gianarb:
Ok, I see I still don't really understand the full picture, I hope the diagram and a concrete use case will help me!

mmlb:
@gianarb I've added a diagram, please let me know if that helps or if needs something else to help.

0015/README.md
gianarb: That's the way I understood that, but I am open to learning more about what hardware means.

mmlb: Yeah we don't have any term internally for these things, its just the "site prov" machines or the actual host names, but nothing generic. :thinking: we should maybe figure one out.

mmlb: oh I think I've heard "remote-worker" vs "local-worker" earlier on in the design/beginnings of tink. That might be good to use here?

0015/README.md
```diff
@@ -0,0 +1,153 @@
+---
+id: 0015
+title: HTTP Routing For Services
+status: ideation
+authors: Manny <manny@packet.com>
+---
+
+## Summary
+
+HTTP routing for services started by tink-worker running on "provisioner" machines.
```
displague: As a potential reviewer, I need more context here so my comments will be a bit meta and vague.

How does this relate to #18, is #18 a prerequisite? If so, what features of #18 are needed. How will this proposal change that behavior.

displague: As a potential reviewer, I need more context here so my comments will be a bit meta and vague.

How does this relate to #18, is #18 a prerequisite? If so, what features of #18 are needed. How will this proposal change that behavior?

displague: As a potential reviewer, I need more context here so my comments will be a bit meta and vague.

How does this relate to #18, is #18 a prerequisite? If so, what features of #18 are needed. How will this proposal change that behavior?

mmlb: Opposite, this a prerequisite for #18. I think we should start the discussion over there.

gianarb:
Since #21 we moved all the proposals inside the `proposals` directory. Do you mind to rebase and move your dir?
Thanks

mmlb:
> Since #21 we moved all the proposals inside the `proposals` directory. Do you mind to rebase and move your dir?
> Thanks

done

tstromberg:
@gianarb - can you approve/merge this PR?

gianarb (approved):

PR #21, opened 2020-09-25
Created a gen cmd that helps to generate a summarized html website

gianarb:
A missing point of the proposal 0001 was to have a generated website
that act as a summarization of the proposals.

This PR contains a very simple `gen` CLI tool that generates an HTML
page with the list of the current merged proposal and their status.

Screenshot 2020-09-25 at 10 28 26

TODO: 

- [ ] Set up a GitHub action that published this website

PR #22, opened 2020-09-28
0016: Default worklow behaviour

thebsdbox:
## Description

This proposal defines the specifics for running a default workflow for unknown hardware

## Why is this needed

This is the beginning work for projects such as DCIM, or assisting in DataCentre OPS

### DIFF
--- a/proposals/0016/README.md
+++ b/proposals/0016/README.md
@@ -1,0 +1,35 @@
+---
+id: 0016
+title: Propose Default workflow
+status: ideation
+authors: Dan <daniel.finneran@gmail.com>
+---
+
+## Summary
+
+This additional feature would allow an end user to specify a workflow to be used against a machine that has no `hardware` profile specified within Tinkerbell. The main feature that this would enable would be for DCIM or hardware discovery where any new unknown machines are automatically prossed by a specified workflow that could perform actions such as:
+
+ - Secure wipe
+ - Forensic analysis
+ - DataCentre inventory
+ - Hardware capture
+
+
+This could ultimately lead to the capability to build pools of hardware (grouped by a hardware profile (GPUs/SSDs/CPU type)) that we could start to provide "cloud like" functionality around.
+
+## Goals
+
+1. Specify a default workflow
+2. If Tinkerbell doesn't "know" the hardware then assign that workflow
+
+## Not Goals
+
+N/A
+
+## Content
+
+The same functionality was implemented in `plunder` in order to build a pool of discovered hardware, which the `cluster-api` provider could allocate unused hardware when provisioning was required. 
+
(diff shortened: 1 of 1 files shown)


gianarb:
This is a lack of knowledge from my side but I always thought Tinkerbell had two different way of working.
The one I know and we have in the documentation requires each hardware to be registered. There is another way that should work differently, I assume like: "hardware not registered yet is capable of registering itself". If that's the case maybe some of this work is already done

gianarb:
Since #21 we moved all the proposals inside the `proposals` directory. Do you mind to rebase and move your dir?
Thanks

mmlb:
Tinkerbell does not have this built-in currently, but boots has something like what this proposal is ... proposing :D. Just not as automated, like at all.

We have recently landed the "Hardware Capture" part using https://github.com/packethost/discover-metal into OSIE in https://github.com/tinkerbell/osie/pull/150. Having tink default to a workflow that runs this image was part of the future plan already :D.

0016/README.md
```diff
@@ -0,0 +1,35 @@
...
+authors: Dan <daniel.finneran@gmail.com>
+---
+
+## Summary
+
+This additional feature would allow an end user to specify a workflow to be used against a machine that has no `hardware` profile specified within Tinkerbell. The main feature that this would enable would be for DCIM or hardware discovery where any new unknown machines are automatically prossed by a specified workflow that could perform actions such as:
+
+ - Secure wipe
+ - Forensic analysis
+ - DataCentre inventory
```
mmlb: What kind of DC inventory are you thinking? Presumably an untrusted machine shouldn't have network access to all parts of a DC.

gianarb:
Ok now that we know this feature is not there I think we should merge and accept this issue. @thebsdbox do you mind to do a  review of the questions, update status to accepted and move to the right directory?

gianarb (changes requested):
Ok now that we know this feature is not there I think we should merge and accept this issue. @thebsdbox do you mind to do a review of the questions, update status to accepted and move to the right directory?

0016/README.md
thebsdbox: The Datacentre inventory would just be whatever tinker bell is capturing from the machines that boot the first time when we don't know about them.

- NICs
- CPU
- Memory 
- GPU
- Disk config
- Is anything installed on the servers
- Firmwares?

mmlb: ah well that doesn't sound like DC inventory, more like node inventory right? Just a wording that might be confusing.

mmlb (dismissed):

gianarb (dismissed):

gianarb (approved):

PR #23, opened 2020-10-03
Initial README

gianarb:
When I open this repository I feel a bit sad about not having a README,
but it is also a lot of work to write a good one. For now the proposal
001 is the best way we have I think to understand how this repository
works.


thebsdbox (approved):
Yeah, 

I think as an initial place for contributors to head too.. we need some information about how the proposals system works.

gianarb:
@thebsdbox did you read the PR? There is a link to the 0001 proposal that contains how it works.

PR #25, opened 2020-10-13
0018: Propose Instance object

mmlb:
## Description

Propose an Instance object (name up for consideration)

-----
[View rendered proposals/0018/README.md](https://github.com/mmlb/proposals/blob/0018/proposals/0018/README.md)

### DIFF
--- a/proposals/0018/README.md
+++ b/proposals/0018/README.md
@@ -1,0 +1,191 @@
+---
+id: 0018
+title: Add Instance Object Definition
+status: ideation
+authors: Manny <mmendez@equinix.com>
+---
+
+## Summary
+
+The current hardware data model does not support any OS installation details.
+Installing an OS is one of the main focus areas of the Tinkerbell project at the moment.
+This exists in the legacy Packet Platform under "instance" in the hardware object and a similar concept is in [metal3].
+
+This would be used by Boots and Hegel (at least) to serve their needs.
+
+## Goals and not Goals
+
+We have a goal of keeping the Hardware data relatively static.
+We need a locaion to place instance/provision/installation-info that changes from provision to provision.
+It is not a goal to stick this info in `elastic_data`.
+
+## Content
+
+This proposal is about adding a storage concept for data that is not as permanent as Hardware, and that can persist across multiple Workloads.
+This would mimic the EquinixMetal life-cycle of a machine, provisioned and thus occupied until the customer deletes and the machine is deprovisioned.
+This life-cycle is also useful in non EM use cases as it allows a worflow to end and a future workflow to act upon the previous os intallation state.
+
+#### Hegel
+
+Hegel currently serves up EC2 style metadata by transforming the JSON blob in hardware.metadata to the compatible form.
+The JSON blob in Hardware.metadata is not specified in order to be flexible to non EquinixMetal operators, yet Hegel's transformation logic is fixed.
+So in reality this is just an undocumented JSON blob that is actually very useful to the project overall.
+
+#### Boots
+
+Boots is currently dynamically creating the iPXE script from data in Hardware.
+We generally want to keep the Hardware data static.
+But we would end up needing to modifying it to add ephemeral info, such as IPs (public/private v4 and public v6 as EquinixMetal does today), hostname, os installer to use...
+
+## APIs
+
+The API consists of defining a minimal and extensible Instance object as a field in Hardware, via Protobufs.
+We already have a protobuf file in tree ([packet]) that defines the non-tinkerbell Instance JSON object as Protobuf.
+This will be used as the starting point, moving fields that are unnecessary for Tinkerbell/Hegel/Boots into a meta field.
+If interest builds we can hoist meta values into the Instance protobuf definition.
+
+```protobuf
+syntax = "proto3";
+
+option go_package = "packet";
+
+package github.com.tinkerbell.tink.protos.packet;
+
+message Instance {
+  string id = 1;
+  string state = 2;
+  string hostname = 3;
+
+  // fork in the road! choose a path
+  // path 1Left
+  message CustomIPXE {
+    string url = 1;
+    bool always_pxe = 2;
+  }
+  message OperatingSystem {
+    string distro = 1;
+    string version = 2;
+    string tag = 3; // snapshot? revision? ...?
+  }
+  oneof OS {
+    CustomIPXE custom_ipxe = 4;
+    OperatingSystem operating_system = 5;
+  }
+  // path 1Center
+  message OperatingSystem {
+    string distro = 1;
+    string version = 2;
+    string tag = 3; // snapshot? revision? ...?
+    bytes extra = 4;
+  }
+  // path 1Right
+  message OperatingSystem {
+    string distro = 1;
+    string version = 2;
+    string tag = 3; // snapshot? revision? ...?
+    string ipxe_url = 4;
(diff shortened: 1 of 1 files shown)


gianarb:
Reading the proposal I had an intuition:

> We need a location to place instance/provision/installation-info that changes from provision to provision.

Instance to me looks a bit too generic. An instance of what? Obviously, it is the concretization of hardware at some specific point in time (provision to provision) I get that, but I am wondering if there is a more specific term.

Other than that the proposal looks good to me.

mmlb:
Yeah the name is very much up for options. Installation? Just OS...?

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+Path 1Center is a bit of a mix, with benefits and warts from both sides.
+The common/well specified options are well specified, but leaves the generic bits out for interpretation.
+
+I lean towards trying out the 1Left path.
+
+#### Fork 2: To bitflags or not bitflags?
+
+Should we do bitflags in Instance.OperatingSystem.IP.flags (Path 2A) or break them out to bool `public` and `management` like currently done in [protos/packet][packet] (Path 2B)?
+
+I don't particularly like "exploding" flags into fields, but that could just be my once-upon-a-time embedded developer thinking leaking.
```
detiber, line 155: My personal opinion here is that we should start with fields and only optimize into bitflags if/when it's needed.

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+message Instance {
+  string id = 1;
+  string state = 2;
+  string hostname = 3;
+
+  // fork in the road! choose a path
+  // path 1Left
+  message CustomIPXE {
+    string url = 1;
```
detiber, line 62: Should this also have a field for an ipxe script as well and not just a URL?

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+    bool always_pxe = 2;
+  }
+  message OperatingSystem {
+    string distro = 1;
+    string version = 2;
+    string tag = 3; // snapshot? revision? ...?
+  }
+  oneof OS {
+    CustomIPXE custom_ipxe = 4;
+    OperatingSystem operating_system = 5;
```
detiber, line 72: Should there also be a third option to specify kernel/initrd images directly?

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+    Family family = 4;
+
+    // fork in the road! choose a path
+    // path 2Left
+    enum Flags {
+      NONE = 0;
+      PUBLIC = 1;
+      MANAGEMENT = 2;
+    };
+    uint32 flags = 5;  // bitwise-or of Flags
```
detiber, line 110: I'm wondering if this should be something that is defined separately and not hardcoded. I can potentially see users wanting to have more granular network segregation than just public and management networks.

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+    // vs
+    //path 2Right
+    bool public = 5;
+    bool management = 6;
+
+  }
+  repeated IP ips = 6;
+
+  string userdata = 7;
```
detiber, line 120: Forgive my lack of understanding here, but is userdata here meant as a way to provide automated injection into Hegel for it later to be consumed by the instance at boot time? 

If so, it might also be good to provide an extra abstraction around userdata so that we could eventually provide higher levels of granularity around permissions to access userdata compared to standard instance data. This is something that has consistently been a challenge across all cloud providers (and other api-driven infastructure, such as vSphere), we have the potential to avoid perpetuating it here.

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+  string crypted_root_password = 8;
```
detiber, line 121: Is there any need to worry about potential conflicts between this field and userdata?

It would also be nice to avoid storing any passwords here, for a similar reason as mentioned above about userdata.

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+  repeated string tags = 9;
+  repeated string ssh_keys = 10;
```
detiber, line 124: Is there any need to worry about potential conflicts between this field and userdata?

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+```
+Where the code will just accumulate the flags seen.
+
+This is a bit more formal of a definition of bitflags than "enums + int" but seems weird to me, I can see this being used though.
+This path doesn't protect against repeated dupes, but I doubt that would be a big deal.
+
+#### Extensions?
+
+I can already think of a couple of extensions that we'd need for EquinixMetal:
+* Licensing fields for some OSes
```
detiber, line 174: Could this be handled through composition?
```sh
  message OperatingSystem {
    string distro = 1;
    string version = 2;
    string tag = 3; // snapshot? revision? ...?
  }
  message LicensedOperatingSystem {
    OperatingSystem os = 1;
    string license = 2;
  }
```

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+  // fork in the road! choose a path
+  // path 1Left
+  message CustomIPXE {
+    string url = 1;
+    bool always_pxe = 2;
+  }
+  message OperatingSystem {
+    string distro = 1;
+    string version = 2;
+    string tag = 3; // snapshot? revision? ...?
```
detiber, line 68: Are these two fields going to be sufficient to enumerate the different types of versioning practices for various OSes? Or will it lead to trying to shoehorn everything into the model we provide?

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+This path doesn't protect against repeated dupes, but I doubt that would be a big deal.
+
+#### Extensions?
+
+I can already think of a couple of extensions that we'd need for EquinixMetal:
+* Licensing fields for some OSes
+* Spot Market related fields
+* IQN, used for block storage
+
+And the 2 ways I can think of to handle the extensions:
```
detiber, line 178: I think it really depends on how the extensions are expected to be used/consumed.

If the goal is to only enable the ability to specify extra configuration for external (not a part of Tinkerbell) services, then I think the approaches mentioned below make sense. However, if we expect internal services to make use of the various extensions, then I think it might be better to encode them in through protobuf composition or make use of compatible types to provide a hacky approach to inheritance.

proposals/0018/README.md
```diff
@@ -0,0 +1,191 @@
...
+## Goals and not Goals
+
+We have a goal of keeping the Hardware data relatively static.
+We need a locaion to place instance/provision/installation-info that changes from provision to provision.
+It is not a goal to stick this info in `elastic_data`.
+
+## Content
+
+This proposal is about adding a storage concept for data that is not as permanent as Hardware, and that can persist across multiple Workloads.
+This would mimic the EquinixMetal life-cycle of a machine, provisioned and thus occupied until the customer deletes and the machine is deprovisioned.
```
detiber, line 25: I think having more information about the expected lifecycle of the proposed "instance" type would help.

A few questions that I have:
- Is the Instance expected to be user-created, created through a workflow, or other automated means?
- How is the Instance linked to underlying hardware?
- How is appropriate hardware chosen for an instance?

mmlb:
This one too @gianarb, thanks.

gianarb (approved):

PR #30, opened 2020-10-30
0019: Propose PBnJ.next

jacobweinstock:
## Description

This proposal defines the update to the PBnJ service.

---
view rendered proposal [0019: Propose PBnJ.next](https://github.com/jacobweinstock/proposals/tree/pbnj.next/proposals/0019)


gianarb:
This proposal to me does not sound very actionable :D It speaks about a more: resilient, scalable, robust, reusable program. Who is that crazy to say no!? Let's do it, please! 📅  My suggestion here is, I think we should merge this PR as it is, and you can open another one with status moved to `discussion` and eventually answers to some of the questions below. Is it ok for you?

But, I can't tell you a few things that I do not appreciate about the current implementation. I get what it does but I don't know when it is useful for me, which BMC it supports? How can I hook it to Tinkerbell?

The diagram contains a lot of terms that I don't understand. I presume it describes two different feature PBnJ supports: power on a server and executing a task (which tasks?). The gRPC server is the same as well as all the other components? Are all of them part of PBnJ?

Did you speak about the in-memory store for tasks that makes this system not that scalable, why? From where those tasks are coming from? In the proposal there is no mention of an alternative, will it stay an in-memory datastore if not what it will become?

I hope all those questions can help the community to build some context around this proposal and PBnJ itself that even for us in DevRel is kind of "what is it?"
Thanks

jacobweinstock:
Hey @gianarb, thank you for the feedback! Sorry for the poorly articulated proposal. 

> My suggestion here is, I think we should merge this PR as it is, and you can open another one with status moved to discussion and eventually answers to some of the questions below. Is it ok for you?

Sounds good to me. I will answer all questions and make it better!

gianarb (approved):

PR #31, opened 2020-11-03
0020: Proposal for Template Versioning

parauliya:
Signed-off-by: parauliya <aman@infracloud.io>

-----
[View rendered proposals/0020/README.md](https://github.com/parauliya/proposals/blob/template_versioning/proposals/0020/README.md)

### DIFF
--- a/proposals/0020/README.md
+++ b/proposals/0020/README.md
@@ -1,0 +1,157 @@
+---
+id: 0020
+title: Template Versioning
+status: ideation
+authors: Aman Parauliya <aman@infracloud.io>
+---
+
+## Summary
+
+As per the proposal [0013], we are planning to induce workflow service as a part of the template which is used to create a workflow.
+
+As we add new features, we may have to modify the existing template format. Therefore, from a future perspective, we should support template versioning so that we can support existing template formats.
+
+## Goals and not Goals
+
+### Goals
+
+Provide template versioning so that current format or schema of the template should be compatible as well as with the new one.
+
+### Not Goals
+
+Current template schema should be depricated.
+
+## Content
+
+This feature will make the life of existing user a bit easier because of the following points:
+
+- User doesn't need to migrate to new template schema since we have the support of existing templates.
+- User will also have the flexibility of using the kind of template he/she would like to use or in another words whichever template schema is fit for his/her usecase.
+
+## Lifecycle
+
+With each of the release we can have a different version of template if template schema has been modified in that particular release. And based on the versions we can have different structure/functions in our code to work with different formats/schemas of a template which is a very important because actual workflow lies withtin the template itself.
+
+Following is the example of two different template schemas:
+
+### 1. Current template
+
+```yaml
+version: "0.1"
+name: ubuntu_provisioning
+global_timeout: 6000
+tasks:
+  - name: os-installation
+    worker: {{.device_1}}
+    volumes:
+      - /dev:/dev
+      - /dev/console:/dev/console
+      - /lib/firmware:/lib/firmware:ro
+    environment:
+      MIRROR_HOST: <MIRROR_HOST_IP>
+    actions:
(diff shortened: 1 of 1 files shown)


proposals/0020/README.md
```diff
@@ -0,0 +1,157 @@
...
+        Name          string `yaml:"name"`
+        ID            string `yaml:"id"`
+        GlobalTimeout int    `yaml:"global_timeout"`
+        Services      []Service `yaml:"services"`
+        Tasks         []Task `yaml:"tasks"`
+}
+
+```
+
+Now if we would like to support both of the above templates then we need to identify the struct in which we need to parse the template which we can identify based on the versions.
```
detiber, line 149: Kubernetes handles this by having a storage version (that is rich enough to store data about all supported versions) and allowing the user to specify which version they are providing, we can likely do similar here, and since from what I can tell the changes appear purely additive in nature, we could also use the new version as the storage version.

gianarb:
I don't think I understand this proposal. Kubernetes has its own way of versioning and managing versions. Very complex, probably very well done looking at how many extensions and CRD developers run successfully.

```yaml
version: "0.1"
name: hello_world_workflow
global_timeout: 600
tasks:
  - name: "hello world"
    worker: "{{.device_1}}"
    actions:
      - name: "hello_world"
        image: hello-world
        timeout: 60
```
This is our template TODAY. it is in YAML format, and it has a `version` field already. This is not used probably, but it is there as a safeguard as soon as we will need a bc break.

Tinkerbell is currently not stable and BC breaks are expected. Proposal [0013](https://github.com/tinkerbell/proposals/pull/18) is not a BC break, it is an additive proposal and what is working today will work when 0013 will become reality.

I think it is a premature optimization to implement the code path that multiple versions running at the same time requires.

This sounds like a reasonable proposal only if versioning was not planned at all. But there is from a user point of view the `version` field already, so for now it looks enough for me

gianarb:
I think this is not valid anymore @parauliya since @gauravgahlot added the revision to a template. Do you mind changing status to abandoned specified this PR  https://github.com/tinkerbell/tink/pull/438? 

Thanks a lot

gauravgahlot:
@gianarb Maybe this is still relevant wrt [tink-issue/440](https://github.com/tinkerbell/tink/issues/440). However, it may require some update and clarity. 

@parauliya Could you please check what's missing here?

parauliya:
@gauravgahlot I think the issue which [tink-issue/440](https://github.com/tinkerbell/tink/issues/440) is different but yes as per the initial thoughts by @mmlb it seems like versioning is required.
@gianarb This proposal is different from the PR [tinkerbell/tink#438](https://github.com/tinkerbell/tink/pull/438). In the PR we are talking about maintaining internal revisions of templates but this proposal talks about the version provided by the user inside the template which can define the schema of template. There can be different schemas for different version and based on that different validation may apply on the templates before creating them.

parauliya:
@gianarb @mmlb What do you think about this proposal now? I mean is it make sense to take it ahead or should we just abandoned it?

tstromberg:
@detiber @gianarb - do you mind giving this proposal another look? 

I'm new here, but I'm going to be honest here and say I don't quite understand what this PR provides versus the current data structure.

gianarb:
I think the data structure is wrong as well. The data model that Tinkerbell uses has a lot of things I do not understand 👍

nshalman:
> I think the data structure is wrong as well. The data model that Tinkerbell uses has a lot of things I do not understand 👍

@gianarb should we keep this open or should we close it?

gianarb:
For me it can be marked as aborted and merged

jacobweinstock:
@parauliya do you want to update this PR to `status: ideation` to `status: aborted`? or should we just close this PR?

gauravgahlot (approved):
The status has been changed to `aborted`.

PR #33, opened 2020-11-13
Move 0007 to committed phase

gianarb:
We have sandbox now, up and running. I think we can move this proposal
to `committed`

-----
[View rendered proposals/0007/README.md](https://github.com/tinkerbell/proposals/blob/move0007-committed/proposals/0007/README.md)


thebsdbox (approved):
Agreed! Lets move this to `committed`

PR #34, opened 2020-11-13
0021: The Tinkerbell Action Hub

gianarb:
-----
[View rendered proposals/0021/README.md](https://github.com/tinkerbell/proposals/blob/0021-initialize/proposals/0021/README.md)


thebsdbox (approved):
I'm definitely in favour of having an actions hub!! 

Would we consider a `workflows` hub or just leave that to end users?

PR #35, opened 2020-11-17
0019 PBnJ.next Proposal Discussion

jacobweinstock:
## Description

Updated proposal to hopefully be more succinct about its
purpose and place in the Tinkerbell stack.

view rendered proposal 0019: [Propose PBnJ.next](https://github.com/jacobweinstock/proposals/blob/0019-pbnj.next/proposals/0019/README.md)

## Why is this needed

Make the proposal clearer.

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

-----
[View rendered proposals/0019/README.md](https://github.com/jacobweinstock/proposals/blob/0019-pbnj.next/proposals/0019/README.md)


gianarb:
Thank you for your explanation, do you imagine PBnj as a standalone project that the user should integrate by itself or it will be used by the tink-server to keep a single point of contact from the user point of view?

jacobweinstock:
> Thank you for your explanation, do you imagine PBnj as a standalone project that the user should integrate by itself or it will be used by the tink-server to keep a single point of contact from the user point of view?

Personally, I think it should be hooked into tink.

gianarb:
So just as confirmation, in your option, the end-user should not talk to PBNJ normally. Everything should go throw the tink-server gRPC API. Right?

jacobweinstock:
> So just as confirmation, in your option, the end-user should not talk to PBNJ normally. Everything should go throw the tink-server gRPC API. Right?

Yes, that's my opinion. What do you think?

gianarb:
I think the same as well 👍  Are we brave enough to phrase that as part of the proposal or is it out of scope?

jacobweinstock:
I think we could mention it. More than that though feels out of scope for this proposal. 

Something like this?
"While PBnJ does not currently have any integration into the Tinkerbell stack, it is very likely that it will be integrated into tink-server and be used to provide additional functionality to workflows in the future."

gianarb (dismissed):
For me, we can change the status to approved and we can marge!

proposals/0019/README.md
```diff
@@ -1,66 +1,58 @@
...
+PBnJ is a Restful API service that is tasked with interacting with Base Management Controllers (BMC). Among many other things, these BMC's can power a machine on and off and set the next boot device to PXE, BIOS, disk, etc. As part of a Tinkerbell workflow execution, the first step is
 
-This proposal is intended to provide an option for how to increase the
-maintainability, robustness and extensibility of the service.  
+> On the first boot, the Worker is PXE booted
+
+This is where PBnJ could be called to set the next boot device to PXE and then called again to reboot the machine. PBnJ currently support BMC's that work with the `ipmitool` or `racadm` tool. The following vendors are known to work: Supermicro, Dell, ASRockRack.
+
+Currently, there is no mechanism in Tinkerbell to hook into this functionality (unfortunately, that mechanism is not what this proposal is about). The PBnJ service is only provided as a convenience for End-Users to hook into. While PBnJ does not currently have any integration into the Tinkerbell stack, it is very likely that it will be integrated into tink-server and be used to provide additional functionality to workflows in the future.
```
mmlb: tink-server will not need any support for pbnj. All workflows will start with a couple of pbnj actions.

mmlb:
Ooops didn't see this convo here, going to restate my in-line comment.

No we shouldn't have tink server support PBnJ. There's no need for it. We can easily have a workflow action that does the bmc work.
The only support we'd need from tink-server is of secrets management (for bmc creds). We should have a way to flag some bits of the hardware as secret and only allow it to certain workers. I think we have some mentions somewhere of "privileged" workers for this. Having support for secrets and privileged workers is more generic and thus more useful too.

jacobweinstock:
> Ooops didn't see this convo here, going to restate my in-line comment.
> 
> No we shouldn't have tink server support PBnJ. There's no need for it. We can easily have a workflow action that does the bmc work.
> The only support we'd need from tink-server is of secrets management (for bmc creds). We should have a way to flag some bits of the hardware as secret and only allow it to certain workers. I think we have some mentions somewhere of "privileged" workers for this. Having support for secrets and privileged workers is more generic and thus more useful too.

Hey @mmlb , I've updated the wording to be more generic with future predictions of PBnJ's integration into the tinkerbell stack. Let me know what you think. Would like to see this merged. Thanks!

jacobweinstock:
Hey @gianarb , made some changes. mind reviewing again, please?

gianarb (approved):
it is ready to go!! Thanks

PR #36, opened 2020-12-15
0021: move to discussion

gianarb:
Moved the proposal to the discussion and I added more about how we can implement the Action Hub via CNCF ArtifactHub

-----
[View rendered proposals/0021/README.md](https://github.com/tinkerbell/proposals/blob/0021/discuss/proposals/0021/README.md)

### DIFF
--- a/proposals/0021/README.md
+++ b/proposals/0021/README.md
@@ -1,7 +1,7 @@
 ---
 id: 0021
 title: The Tinkerbell Action Hub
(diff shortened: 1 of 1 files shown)


thomcrowe (approved):

gianarb:
:P @thomcrowe made a mistake, but the real mistake is from us not taking care of access to repositories :)

PR #46, opened 2021-08-27
Add support for the Kubernetes Resource Model

micahhausler:
## Description

Added proposal 0026 Support for the Kubernetes Resource Model.

-----
[View rendered proposals/0026/README.md](https://github.com/micahhausler/proposals/blob/0025/proposals/0026/README.md)


rawkode:
Rather than needing a fully fledged Kubernetes cluster, it would be cool to see this work with kcp

https://github.com/kcp-dev/kcp

micahhausler:
> Rather than needing a fully fledged Kubernetes cluster, it would be cool to see this work with kcp
> 
> https://github.com/kcp-dev/kcp

That would be super cool. As long as KCP supports [CRDs](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/), [admission webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/
), [finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/), [watches](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes), [leases](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/lease-v1/), and [RBAC](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) (which I think it does support all of those), it should be portable.

proposals/0025/README.md
```diff
@@ -0,0 +1,199 @@
+---
+id: 0025
+title: Tinkerbell on Kubernetes
```
tstromberg: Can you rename this something specific, such as `Support for the Kubernetes Resource Model' or 'Support for provisioning via Kubernetes API's"?

Tinkerbell can already run on Kubernetes, which makes this proposal confusing. What you seem to be proposing is buying fully into its resource model:  https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/resource-management.md - and allowing machines to be provisioned using the Kubernetes API.

proposals/0025/README.md
```diff
@@ -0,0 +1,199 @@
...
+* Kubernetes natively supports streaming watches for CRD types.
+  Any process (PBnJ, Tink API, CAPT, etc) that need streaming updates can leverage Kubernetes Watches.
+* Multi-worker workflows will require some backend changes, but that will be opaque to Tink worker clients.
+* The only authorization mode required for the Tinkerbell API will be to support
+  * All Kubernetes clients (Boots, Hegel, PBnJ, etc) can have least-permission RBAC policies to limit the permissions for each respective client.
+* High-availability Kubernetes and etcd can be delegated to a cloud provider or an on-premise Kubernetes cluster
+
+### Tradeoffs
+
+TODO
```
tstromberg: You may want to mention something about the bootstrapping/circular dependency issue here, as often the nodes of a Kubernetes cluster would be provisioned by a system like Tinkerbell.

The two cases to think about in particular are initial setup and emergency provisioning during an outage.

proposals/0025/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+A validating webhook will be used to validate Templates and Workflows as they are created.
+It will not require any Kubernetes API access, but needs to be reachable by the Kubernetes API server.
+
+**Tinkerbell Administrator**
+
+Tinkerbell administrators might require varying levels of access in order to perform CRUD operations on any or all Tinkerbell types.
+Kubernetes cluster administrators can define custom levels of access with RBAC policies to grant Tinkerbell admins the correct least-privilege level of access.
+
+## Alternatives
```
tstromberg: One thing to consider in this doc is how to safely transition existing users of Tinkerbell to this model. We can't suddenly cut-over to the new API and break existing users. Similar to Kubernetes itself, we probably need to hide this new feature work behind a flag for some time: particularly if the previous behavior is to be deprecated.

tstromberg (changes requested):

proposals/0025/README.md
micahhausler: Thanks! Good callout, I renamed the title

proposals/0025/README.md
micahhausler: I added a section on Administration and addressed bootstrapping and backup/restore.

proposals/0025/README.md
micahhausler: I added a migration section

andy-v-h:
I like that idea of creating in tree k8s support. Specifically making it easier to do a hitless reload on `hegel` by updating a ConfigMap, for example. However, I'm not sure a CRD based approach is the direction we would be best. My primary concern is the additional load on `etcd` and resiliency there. A scenario I find concerning is where you have relatively small amounts of compute available for running tinkerbell, and requiring it to rely on `etcd` vs being able to use an external DB service. And in a DR scenario, how would a recovered tinkerbell not reprov all of my hosts?

micahhausler:
> I like that idea of creating in tree k8s support. Specifically making it easier to do a hitless reload on hegel by updating a ConfigMap, for example. However, I'm not sure a CRD based approach is the direction we would be best. My primary concern is the additional load on etcd and resiliency there. 

ConfigMaps in Kubernetes are unstructured and don't allow for strong typing. This causes lots of practical validation problems. CRDs are roughly equivalent in terms of efficiency and load, and they allow for much easier client development. etcd specifically can be operated in many different modes, from single node (lower resiliency) more typically 3 nodes for increased availability. Neither etcd nor Kubernetes need to be operated on the same compute that Tinkerbell is operated on. This proposal is agnostic to where Kubernetes is run, and operators can connect Tinkerbell to Kubernetes in a managed cloud environment where compute resource is less of a concern.

> A scenario I find concerning is where you have relatively small amounts of compute available for running tinkerbell, and requiring it to rely on etcd vs being able to use an external DB service. 

Neither Kubernetes nor etcd would be required to run on the same compute as Tinkerbell. This would be equivalent to connecting to an external DB service.

> And in a DR scenario, how would a recovered tinkerbell not reprov all of my hosts?

Check out the **Migrating between Kubernetes clusters, Backups, and Restoration** section under [Administrative Tasks](https://github.com/micahhausler/proposals/blob/0025/proposals/0025/README.md#administrative-tasks) but the short version is that the `.status` of Workflows would be included in backups and restoration and machines would not be re-provisioned.

Does that address your concerns?

splaspood:
I would love to see tinkerbell support this as one of many possible data backends, forcing us to create a well defined interface between tink server and n number of possible underlying data stores.

displague:
Like @splaspood, I wonder what the tinkerbell services would look like if the backend could be toggled between a local filesystem (`manifest.d/` style directories), postgres, and Kubernetes resources.

In order to support these three different storage models, a pluggable data interface would benefit from a Tinkerbell data event system.

Rather than having tinkerbell poll a database, or watch a directory, or watch a set of kubernetes CRs and configmaps, tinkerbell could await data events while the data interface modules take on the polling and watching.

Does this mean that tinkerbell takes on another grpc service? Perhaps.
Does tinkerbell need caching or relationship mapping between resources? Maybe? (I don't think so.)

What events would tinkerbell need to listen for from pluggable data modules?
What features does tinkerbell need to update data stored in generic data modules?

displague:
Reconsidering my last comment, rather than invent a new format of event drivers and data passing (which are some of the problems we are trying to solve with this proposal), by taking the K8s API as the data and eventing backend (not necessarily with containers involved) we inherit a standard and well-evolved set of tools and opinions that bring stability (to the API and data backend). 

We solve the pluggable data backend because we can use services like https://github.com/k3s-io/kine to provide a postgresql, MySQL, and yet-to-be-added backends.

micahhausler:
> Reconsidering my last comment, rather than invent a new format of event drivers and data passing (which are some of the problems we are trying to solve with this proposal), by taking the K8s API as the data and eventing backend (not necessarily with containers involved) we inherit a standard and well-evolved set of tools and opinions that bring stability (to the API and data backend).
> 
> We solve the pluggable data backend because we can use services like https://github.com/k3s-io/kine to provide a postgresql, MySQL, and yet-to-be-added backends.

YES! This is exactly it!

jacobweinstock:
This is a great proposal and the WIP demo at the community meeting was really great. I do like the idea of putting this behind a flag.

micahhausler:
Here's the current code I demoed in the community meeting today:
* Sandbox: https://github.com/micahhausler/sandbox/tree/krm
* Boots: https://github.com/micahhausler/boots/tree/krm-feat
* Tink: https://github.com/micahhausler/tink/tree/feature/k8s-mode/

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+  Postgres is a fantastic database, but managing high-availability with graceful failover is not trivial.
+  Using an alternative data store that better supports failure would better help operators to have higher availability and not require downtime for upgrades or failover.
+* **Migrate existing components**.
+  Hegel, Boots, and Tinkerbell API will be modified to use Kubernetes as the datastore
+* **Support migration of existing installations.**
+  Migration tooling and documentation will be provided for existing Workflows, Hardware, and Templates stored in Tinkerbell.
+  Migration should be as straightforward as:
+  * Creating Tinkerbell CRDs in the Kubernetes API
+  * Running a provided migration command
+  * Restarting each Tinkerbell component targeting Kubernetes (ex: `--k8s-mode=true`)
```
displague: This argument supposes that the k8s data backend support will be optional. This asserts that another non-goal would be to remove the existing API interface. 

If we remove that interface, a proxy interface (legacy API -> k8s) could have been proposed, so that is another non-goal. 

This argument also supposes that we will not be dissecting the existing service runtimes from their frontends and data access models. That's a change that I think could be related to this work so it could be enumerated among non-goals.  If we were to do so, we might have a `hegel-pg` and `hegel-k8s` binary, for example, or `hegel-backend-pg` and `hegel-backend-k8s` plugins. But it would appear that today, at least, removing PostgreSQL support is a non-goal.

I would love to see future proposals for both of these changes :-)

displague: This argument supposes that the k8s data backend support will be optional. This asserts that another non-goal would be to remove the existing API interface. 

If we remove that interface, a proxy interface (legacy API -> k8s) could have been proposed, so that is another non-goal. 

This argument also supposes that we will not be dissecting the existing service runtimes from their frontends and data access models. That's a change that I think could be related to this work so it could be enumerated among non-goals.  If we were to do so, we might have a `hegel-pg` and `hegel-k8s` binary, for example, or `hegel-backend-pg` and `hegel-backend-k8s` plugins. But it would appear that today, at least, removing PostgreSQL support is a non-goal.

I would love to see future proposals for both of these changes 🤓

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+* Streaming updates to clients is difficult with the current database
+* Multiple-worker workflows will require numerous API changes
+* There is no authorization in the Tinkerbell API today.
+  In order to support a least-privilege access model, several new authorization modes would need to be supported
+  * An attribute-based authorization method for Tink workers.
+    At the time of writing, any Tink worker is authorized to list all hardware data, including possibly sensitive metadata for other hardware.
+    Today, the Tinkerbell Cluster API (CAPI) provider stores machine CA private key data in Tinkerbell hardware metadata.
+    Ideally, some authorizer that could make decisions based on authentication identity would be used (The Kubernetes analog would be the [Node Authorizer][node-authorizer]).
+    The authenticatoin method in this model could be a network identity, a TPM, or some other bootstrapped authentication credential.
```
displague:
```suggestion
    The authentication method in this model could be a network identity, a TPM, or some other bootstrapped authentication credential.
```

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+  * Either a role-based access method, or some forwarding mechanism for passthrough clients like Boots and Hegel.
+* Operating high-availability Postgres is non-trivial in many environments.
+
+[node-authorizer]: https://kubernetes.io/docs/reference/access-authn-authz/node/
+
+If we only focus on the Tinkerbell API clients as it exists in Q3 2021, the architecture looks something like the following:
+
+[![current-architecture](./tink-arch-1.png)](.tink-arch-2.png)
+
+In this proposal, all non-worker clients of Tinkerbell would become Kubernetes clients.
```
displague:
```suggestion
In this proposal, all non-worker clients of Tinkerbell may become Kubernetes clients.
```

This really comes down to wether or not PostgreSQL support is preserved at all and then whether the user chooses it.

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+In order to provision a machine, the steps would be:
+
+1. User creates a Hardware CRD object in Kubernetes.
+   This is analogous to the current `tink hardware push < hardware.json` command.
+1. User creates a Template CRD object in Kubernetes.
+   This is analogous to the current `tink template create < template.json` command
+1. User creates a Workflow CRD object in Kubernetes.
+   All the user would need to specify in the object `.spec` would be a reference to the Template object, and mappings for devices.
+   This is analogous to the current `tink workflow create -t $TEMPLATE_ID -r "{\"device_1\":\"$MAC_ADDR\"}` command.
+1. The Tinkerbell API would include a Kubernetes workflow controller that would watch for created workflows, and fill out the `.status` with most of the logic that currently exists in the [`CreateWorkflow()`][createwfrpc] RPC.
```
displague: The format of status is not discussed here, but something to consider and then call out if utilized is conditioned statuses, https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

displague: https://github.com/crossplane/crossplane-runtime/issues/198 includes some relatable examples, benefits, and counter-point.

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+    The authenticatoin method in this model could be a network identity, a TPM, or some other bootstrapped authentication credential.
+  * A role-based access method administrative clients.
+  * Either a role-based access method, or some forwarding mechanism for passthrough clients like Boots and Hegel.
+* Operating high-availability Postgres is non-trivial in many environments.
+
+[node-authorizer]: https://kubernetes.io/docs/reference/access-authn-authz/node/
+
+If we only focus on the Tinkerbell API clients as it exists in Q3 2021, the architecture looks something like the following:
+
+[![current-architecture](./tink-arch-1.png)](.tink-arch-2.png)
```
displague, line 95: Tink CRD Controllers are included in the diagram, but not discussed anywhere else. The existing services could be the controllers or the controllers could be some new set of services (including the webhooks) needed to replicate postgresql triggers or inter-service operations.  What were you thinking these would be? Could you provide that somewhere here.

displague: Postgres is shown as the Tink API backend, but no similar backend box is illustrated for the K8s API. To fairly compare the old and new behaviors, should we include an 'etcd' box? That would seem to undercut the role of other services in the Kubernetes API. Perhaps an asterisk could denote that the Kubernetes API represents various services and pluggable data storage models that together provide a _kubernetes in a word_ service?

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+* Multi-worker workflows will require some backend changes, but that will be opaque to Tink worker clients.
+* The only authorization mode required for the Tinkerbell API will be to support
+  * All Kubernetes clients (Boots, Hegel, PBnJ, etc) can have least-permission RBAC policies to limit the permissions for each respective client.
+* High-availability Kubernetes and etcd can be delegated to a cloud provider or an on-premise Kubernetes cluster
+
+### Administrative Tasks
+
+**Migration from Postgres**
+
+In order to support an easier migration from Postgres, we will add a `--kubernetes-mode` or equivalent flag to the Tinkerbell components and for some period of time support two alternate modes: Tinkerbell as it exists today, and the Kubernetes Resource Model.
```
displague: `--kubernetes-mode` was called `--k8s-mode` else where in this proposal.

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+TODO: How long to support both modes?
```
displague: ^ I like this point. Any TODO is best left for 'non-goals', to be sorted later.

It would be nice to get this solved sooner by community consensus, and in the absence of that - this becomes a non-goal.

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+A zero-downtime migration is not in scope for this change mainly because it would add significant complexity and require mirroring data between Postgres and Kubernetes.
+
+**Bootstrapping**
+
+One of the stated design goals is to not require running the Tinkerbell control plane as pods in the Kubernetes cluster.
+Installation of Tinkerbell using Kubernetes as a datastore will differ in the following ways:
+* A Kubernetes cluster will need to exist
+* A Kubernetes [validating admission webhook][validating-webhook] for CRD types will need to be hosted and reachable by the Kubernetes API server.
```
displague: A CRD is sufficient for the external Tink services (now Kubernetes clients) to interact with resources stored in the API.  Users may also interact with the resources without a webhook.

If you remove the controller that watches certain CRD types, the CRDs are still data resources that can be created, updated, and deleted.

What responsibilities do you imagine the validating admission webhook would take on? Why is it needed?

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+  This component is most easily operated as a pod in a cluster, but could be operated and hosted externally.
+  It does not require access to the on-premise Tinkerbell control plane
+* Tinkerbell API, Hegel, and Boots will need network connectivity and credentials to the Kubernetes API.
+  As processes existing outside of the Kubernetes cluster, this will most likely be [x509 client certificates][x509-certs] or exported [Kubernetes Service Account tokens][sa-tokens].
+* Hardware and Templates will be seeded using `kubectl` instead of `tink-cli`
```
displague, line 201: `kubectl` is a proxy for any Kubernetes API client. It may be useful to note that the Kubernetes API is just an HTTPS REST API with OAS3 specs available.

displague: It might help to liken this to the relevant tink-cli commands, and/or SQL verbs.

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+**Bootstrapping**
+
+One of the stated design goals is to not require running the Tinkerbell control plane as pods in the Kubernetes cluster.
+Installation of Tinkerbell using Kubernetes as a datastore will differ in the following ways:
+* A Kubernetes cluster will need to exist
+* A Kubernetes [validating admission webhook][validating-webhook] for CRD types will need to be hosted and reachable by the Kubernetes API server.
+  This component is most easily operated as a pod in a cluster, but could be operated and hosted externally.
+  It does not require access to the on-premise Tinkerbell control plane
+* Tinkerbell API, Hegel, and Boots will need network connectivity and credentials to the Kubernetes API.
```
displague, line 199: It might help to demonstrate what RBAC these services would need (are these read-only clients? Which resources would each service need access to?  Some of this is discussed in the 'Threat Model' and 'User Experience' sections, perhaps link there from here.

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+  As processes existing outside of the Kubernetes cluster, this will most likely be [x509 client certificates][x509-certs] or exported [Kubernetes Service Account tokens][sa-tokens].
+* Hardware and Templates will be seeded using `kubectl` instead of `tink-cli`
+
+Should operators want to run Tinkerbell in an on-premise Kubernetes cluster and use that cluster as the data store, there arises a need to bootstrap the Tinkerbell control plane.
+In order to support this mode, we will need to support migration from either a cloud Kubernetes cluster or a temporarily bootstrap cluster with [`kind`][kind] to a long-lived on-premise management cluster.
```
displague:
```suggestion
In order to support this mode, we will need to support migration from either a cloud Kubernetes cluster or a temporary bootstrap cluster with [`kind`][kind] to a long-lived on-premise management cluster.
```

displague: I don't see the emergence of this mode (moving from kind to another cluster) as necessary for the `bootstrapping` topic.

I think these two sentences could for the intro of your `Migrating between Kubernetes clusters, Backups, and Restoration' section (replacing the text: "previously mentioned bootstrap cluster model" with this pretext).

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+[kind]: https://kind.sigs.k8s.io/
+
+**Migrating between Kubernetes clusters, Backups, and Restoration**
+
+In order to support the previously mentioned bootstrap cluster model, migration from one cluster to another will need to be built.
+Backups of Hardware, Templates, and Workflows can be made using [`kubectl`][kubectl], but simple restoration is not possible as the [`/status` subresource][status-subresource] on CRDs makes the `.status` field a separate API call.
+
+Concerns around data backup and failover were previously delegated to Postgres administration and now become a Kubernetes administration issue.
+
+The `tink-cli` will need to implement backup and restore functionality to support both these administrative concerns.
```
displague: The tink-cli goes away in this proposal, does it not?

If not, what features of tink-cli are maintained?

Perhaps this is a complication of Tinkerbell supporting both a PostgreSQL backend and a Kubernetes backend?

Should `tink-cli` be adapted to work in a Kubernetes mode? Would tink-cli simply not have a TinkAPI to communicate with?

Perhaps the point I am missing is whether or not the TinkAPI persists. Tink API is illustrated to persist in the diagram.  This wasn't obvious to me.  I think Tink API is no longer necessary with this design, unless I am missing some of its benefits or unless it is preserved to continue to support clients like the Tink CLI (and CAPT, for that matter). Is this the intention?

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+[kubectl]: https://kubernetes.io/docs/reference/kubectl/overview/
+[status-subresource]: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource
+
+### Failure modes
+
+**Kubernetes unavailability**
+
+This change introduces a dependency on Kubernetes being available.
+Concerns around data backup and failover were previously delegated to Postgres administration and now become a Kubernetes administration issue.
+As mentioned previously, backup and restoration commands will be added to restore from Kubernetes API unavailability and could help support switching to a different underlying Kubernetes cluster should the primary cluster become unavailable.
```
displague, line 230: 'backup and restore commands will be added' -- could this be considered out of scope? There are numerous Kubernetes API tools for performing such actions, why should the tinkerbell project provide additional tooling? (are you suggesting that there is a need for this?)

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+[k8s-af]: https://k8s.af/
+[kcp]: https://github.com/kcp-dev/kcp/
+[k3s]: https://k3s.io
+
+### User Experience
+
+In order to use Tinkerbell, clients would interact with the Kubernetes API.
+In order to provision a machine, the steps would be:
+
+1. User creates a Hardware CRD object in Kubernetes.
```
displague: Perhaps we could include a more concrete example of how this would be performed:
```yaml
$ kubectl apply -f - <<EOF
apiVersion: 
kind:
metadata:
  name: 
spec:
   ...
EOS
```

displague: or `kubectl apply -f hardware-cr.yaml` to avoid specifics in the proposal.

proposals/0026/README.md
displague: I see there is some explanation of Tink API persistence here: https://github.com/tinkerbell/proposals/pull/46/files?short_path=e8628f6#diff-8f31e22643965ab377983da2fcba8b948c5a61c28744e999adb60c94d8e6fa1bR189-R209

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+[createwfrpc]: https://github.com/tinkerbell/tink/blob/b217be8/grpc-server/workflow.go#L19-L72
+[getwfctxs]: https://github.com/tinkerbell/tink/blob/a56e5cf9/protos/workflow/workflow.proto#L75
+
+## System-context-diagram
+
+See above diagrams.
+
+## APIs
+
+The following [Tinkerbell Workflow APIs][wf-apis] are used by the Tink worker, and would remain.
```
displague: Why can these services not be initially replaced by the Kubernetes model? Do you see them eventually be removed?

Do status conditions, events, logs, and additional CRDs offer the means to move these entities to the Kubernetes API?

.. I think I see.

These are the APIs needed for the provisioning machine to get actions and update state.  That could be more clearly defined.

Why would Tink workers not become Kubernetes clients, for example? Each machine could use a unique service account which would open up more possibilities (potentially reducing the machine threat model).

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+service WorkflowService {
+  rpc GetWorkflowContexts(WorkflowContextRequest) returns (stream WorkflowContext) {}
+  rpc GetWorkflowActions(WorkflowActionsRequest) returns (WorkflowActionList) {}
+  rpc ReportActionStatus(WorkflowActionStatus) returns (Empty) {}
+  rpc GetWorkflowData(GetWorkflowDataRequest) returns (GetWorkflowDataResponse) {}
+  rpc UpdateWorkflowData(UpdateWorkflowDataRequest) returns (Empty) {}
+}
+```
+
+The structure of the Kubernetes CRD types will start out similar to those defined in the [Tinkerbell CAPI Provider][capt-types].
```
displague, line 267: It would be interesting to see in this proposal some description of the choices made in CAPT and how that could change.  Are the resources namespace or cluster scoped, for example, and why?

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+The code defining those types will be migrated into the tinkerbell/tink GitHub repository so other clients could import the Go packages.
+
+[capt-types]: https://github.com/tinkerbell/cluster-api-provider-tink/tree/main/tink/api/v1alpha1
+
+## Threat Model
```
displague, line 273: This section is titled "Threat" Model, but I find that it is also a Data/Service Model overview.

proposals/0026/README.md
displague: Some explanation is given at https://github.com/tinkerbell/proposals/pull/46/files#diff-8f31e22643965ab377983da2fcba8b948c5a61c28744e999adb60c94d8e6fa1bR250

proposals/0026/README.md
displague: I found many of my previous questions in this section.  Perhaps some of the details introduced as Threat Model details could be moved earlier, to where they are first proposed or to a section that broadly covers the service changes and how data access and experience changes (for both services and users).

The Threat model could then be reduced to the perceived attack vectors? Perhaps the threats become a subsection of the experience changes.

For each service:
- how does the data model change (including authorizations needed)
- how are users impacted (old vs new commands)
- what are the new and removed threats

proposals/0026/README.md
```diff
@@ -0,0 +1,275 @@
...
+Tinkerbell administrators might require varying levels of access in order to perform CRUD operations on any or all Tinkerbell types.
+Kubernetes cluster administrators can define custom levels of access with RBAC policies to grant Tinkerbell admins the correct least-privilege level of access.
+
+## Alternatives
+
+The main alternative to achieve the stated design goal around availability would leave the Tinkerbell API alone, and leverage/modify the Tinkerbell server's [internal Database interface][tink-db-iface] to function on top of Kubernetes or a key-value datastore like [etcd][etcd].
+This alternative would help with high-availability deployments, but all the other motivations would remain unaddressed, and need to be implemented in Tinkerbell's API.
+
+Another alternative would be support the current architecture and codebase for existing functionality, and any new functionality would require migrating to use the Kubernetes resource model.
+This proposal does include maintianing the existing functionality and architecture for some period of time, but supporting it indefinitely would add significantly diverging code paths and increased maintenance effort.
```
displague:
```suggestion
This proposal does include maintaining the existing functionality and architecture for some period of time, but supporting it indefinitely would add significantly diverging code paths and increased maintenance effort.
```

This is the answer to the TODO question above: https://github.com/tinkerbell/proposals/pull/46/files#diff-8f31e22643965ab377983da2fcba8b948c5a61c28744e999adb60c94d8e6fa1bR104

Perhaps replace the TODO with this line.

displague (commented):
Thanks for this excellent and well-thought-out proposal, @micahhausler.

I've left some comments as I read through the document, many of the questions that I raise are answered elsewhere. This suggests to me that some information could be reordered for clarity. There were a few details that I felt were missing and I've said as much (this may be hard to gather as I left follow-up comments).

I'd recommend going through this feedback and the previous feedback and make sure that any details that were given in feedback comments and responses have made their way into the proposal.

This proposal, to me, is one of the most significant to be raised. Well done!

proposals/0026/README.md
micahhausler: I intentionally didn't mention etcd, as other K8s layers like K3s may use sqlite or something else. I will make this more clear.

I'll also add notes on the controller process and webhooks

proposals/0026/README.md
micahhausler: Do you want to see an API schema in this proposal? I was thinking it would be better to discuss and iterate on specific API structures in tinkerbell/tink#542

proposals/0026/README.md
micahhausler: I'll remove all specific flag naming from this doc so we can decide on a real name in a PR

proposals/0026/README.md
micahhausler: The webhook can be used for blocking object mutation validation (beyond simple type validation that the CRD gives you), but it will really come into play when we start iterating on API versions in the CRD and need to convert request to the current storage schema

proposals/0026/README.md
micahhausler: The main reason for backup tooling is that a CRD's `.spec` and `.status` are retrieved together, but require separate API calls to update. (`POST /apis/tinkerbell.org/v1alpha1/workflow/<name>` for create, `POST /apis/tinkerbell.org/v1alpha1/workflow/<name>/status` for status), but `kubectl get` works fine for backup.

proposals/0026/README.md
micahhausler: > Why can these services not be initially replaced by the Kubernetes model? Do you see them eventually be removed?

The main reason to only include support for worker APIs is to shorten the migration period and amount of effort required. The ultimate goal is to make all non-worker Tink clients talk to Kubernetes, so instead of a two-step migration where boots/hegel still talk to Tink API, just go directly to the new model. So yes, I do see them eventually removed.

> These are the APIs needed for the provisioning machine to get actions and update state. That could be more clearly defined.

Will do.

> Why would Tink workers not become Kubernetes clients, for example? Each machine could use a unique service account which would open up more possibilities (potentially reducing the machine threat model)
 
I am not inclined to make Tink workers K8s clients at this time. The main reason is I want to limit what an individual (unauthenticated/untrusted) worker can do or see in the API. K8s authz lacks the ability to easily do something like "This worker should only be able to view a specific CRD that matches the worker's name." We basically want attribute based authz, and K8s is all role-based. Yes, you could create a specific RBAC role for each worker, but that gets messy. It's much easier to have a trusted deputy like the Tink API that can correctly and securely propagate the state changes in the storage layer.

proposals/0026/README.md
micahhausler: > This argument supposes that the k8s data backend support will be optional. This asserts that another non-goal would be to remove the existing API interface.

 ~I'll descope PG deprecation for another proposal.~

> This argument also supposes that we will not be dissecting the existing service runtimes from their frontends and data access models. That's a change that I think could be related to this work so it could be enumerated among non-goals. If we were to do so, we might have a hegel-pg and hegel-k8s binary, for example, or hegel-backend-pg and hegel-backend-k8s plugins. But it would appear that today, at least, removing PostgreSQL support is a non-goal.
> I would love to see future proposals for both of these changes :-)

I'll try to articulate this better. The main thrust here is to have two modes of operating Tinkerbell for some period: 
* One as it exists today (architecturally and existing APIs)
* One with the proposed architecture and reduced Tinkerbell API surface. This second mode would be enabled on components (Tink API, boots, hegel) with a feature flag

proposals/0026/README.md
micahhausler: > The tink-cli goes away in this proposal, does it not?
> If not, what features of tink-cli are maintained?
> Perhaps this is a complication of Tinkerbell supporting both a PostgreSQL backend and a Kubernetes backend?

We'll need something that is essentially a special-purpose kubectl to do backup/restore and not a client of the Tink API. Given that it is not a client of Tink API and serves a different function, I'll call it out as something separate.

> Perhaps the point I am missing is whether or not the TinkAPI persists. Tink API is illustrated to persist in the diagram. This wasn't obvious to me. I think Tink API is no longer necessary with this design, unless I am missing some of its benefits or unless it is preserved to continue to support clients like the Tink CLI (and CAPT, for that matter). Is this the intention?

Mentioned elsewhere, but this is essentially to reduce the amount of information a worker can access.

proposals/0026/README.md
micahhausler: >>  This argument supposes that the k8s data backend support will be optional. This asserts that another non-goal would be to remove the existing API interface.

>  ~I'll descope PG deprecation for another proposal.~

I'm going back on this and keeping it in the proposal. I think there are 4 separate stages in this transition.
1. Support full existing API and Postgres backend (Today)
2. Support full existing API and Postgres backend, as well as K8s resource model (KRM)
3. Deprecate most of the existing API and Postgres backend. Continue to build out and support KRM  
4. Remove existing non-worker APIs, only support KRM.

Do we need to define how long stages 2 and 3 last?

proposals/0026/README.md
negz: Speaking of Crossplane, it looks like @displague started on https://github.com/crossplane-contrib/provider-tinkerbell a while back, which would at least provide a KRM compliant shim to Tinkerbell. Perhaps an alternative that wouldn't require rearchitecting the project?

displague: (Going a bit off-topic to consider what this KRM implementation means for a Crossplane Runtime Model (XRM) implementation, and perhaps how they could be the same)

> Speaking of Crossplane,  it looks like @displague started on _a Tinkerbell Crossplane provider_ a while back, which would provide a KRM compliant shim to Tinkerbell

Typically the XRM spec mirrors an external cloud provider API spec as the aforementioned Tinkerell Crossplane Provider does. But if the Tinkerbell spec is implemented as KRM, but does not implement XRM, then a Crossplane Provider would need to interact with Tinkerbell through a  KRM mirroring spec. The cluster could have two nearly identical resources within the same cluster to represent a composable resource.

We could overlook this, as Pods and CronJobs or ReplicaSets have a thin wrapping relationship. 

However, this suggests to me that it should be possible for a controller to implement an XRM compatible interface while adopting the KRM. The benefits of doing so would be that Crossplane compositions (XR) can be applied, managed by Crossplane.  The underlying service, Tinkerbell bare metal orchestration, would transparently serve as a Crossplane Provider.  The Tinkerbell service would not be dependent on Crossplane's installation to add this layer of compatibility.

One complication that comes to mind is the expectation that Crossplane Resources *have a* Provider. In this approach, resources would *be the* provider.  Perhaps, like the Crossplane Helm provider, the provider configuration would refer to the host cluster? Would the default be to report with the local cluster? Would alternatives provider configuration offer connection details for a non-KRM Tinkerbell backplane? Another KRM Tinkerbell backplane?

Adopting Crossplane's packaging mechanisms could provide additional benefits especially for service updates.  Although, in this approach, it's not clear to me what benefits this holds over Helm or Kustomize manifests.

What does this mean in practice? The Tinkerbell KRM implementation (starting with https://github.com/tinkerbell/tink/pull/542)  would adopt XRM types like Referencers, ResourceStatus, ResourceSpec, and ProviderConfig.

https://crossplane.io/docs/v1.4/contributing/provider_development_guide.html#provider-development-guide
https://blog.crossplane.io/crossplane-vs-cloud-infrastructure-addons/ (by @negz) provides the relevant Crossplane context.
https://danielmangum.com/posts/crossplane-infrastructure-llvm/ (by @hasheddan discusses the versatility of the Crossplane resource model)

displague: This is likely too speculative to be actionable. This would be the first controller to act in this XRM compatible way without being first-and-foremost a Crossplane provider.

jacobweinstock (dismissed):
I feel that this has been vetted quite a bit and has been open for comment for a while now. I believe @micahhausler has addressed all comments.

thebsdbox (dismissed):
Awesome, I'd love to see this moved forward.

proposals/0026/README.md
micahhausler: I'll push those specifics into the actual PR reviews

micahhausler:
I've updated the proposal and addressed comments

proposals/0026/README.md
```diff
@@ -0,0 +1,346 @@
...
+* **Require Tinkerbell to be operated as pods inside Kubernetes.**
+  The Kubernetes API will become a dependency of Tinkerbell, but that API could exist in a cloud provider or on-premise.
+* **Require the use of Cluster API ([CAPI][capi]).**
+  The [Tinkerbell CAPI provider][capt] (CAPT) is mentioned in this proposal only for reference of a known Tinkerbell client.
+  Implementation of this proposal will necessitate changes in CAPT, but that is not the core motivation for this proposal.
+* **Implementation of PBnJ as a Kubernetes controller**
+  This can be implemented in a separate proposal.
+* **Make the Tink worker a client of Kubernetes.**
+  Kubernetes doesn't natively support robust attribute-based identities for non-node identities.
+  For example, if the Tink worker were a Kubernetes client, it is not possible with Kubernetes RBAC to ensure that a particular worker can only view or mutate wobernetes does have the [Node Authorizer][norkflows that relate to the worker.
```
displague: **wobernetes** (it looks like you lost a few more words than I can guess)

image

displague: Oh.. **norkflows** too.. I'm going to guess you were writing this line while falling asleep to some classic 80s/90s TV.

image

proposals/0026/README.md
```diff
@@ -0,0 +1,346 @@
...
+In this proposal, all non-worker clients of Tinkerbell will become Kubernetes clients.
+
+[![proposed-architecture](./tink-arch-2.png)](./tink-arch-2.png)
+
+By using the Kubernetes API as the datastore, the motivations are addressed in the following ways:
+
+* Kubernetes natively supports streaming watches for CRD types.
+  Any process (PBnJ, Tink API, CAPT, etc) that need streaming updates can leverage Kubernetes Watches.
+* Multi-worker workflows will require some backend changes, but that will be opaque to Tink worker clients.
+* The only authorization mode required for the Tinkerbell API will be to support
```
displague: "to support ..." this seems cut-off and the sub-bullet seems like an independent statement.

I'm not sure what you were going for here, but perhaps a reword like:


> * The Kubernetes API will manage authorization for all Kubernetes clients (Boots, Hegel, PBnJ, etc). Least-permission RBAC policies will limit the permissions for each respective client.
>  The only authorization mode the Tinkerbell API will need to support/implement is ... (I think this is where you were going 🤷‍♂️ )

proposals/0026/README.md
micahhausler: 😆

micahhausler: You can probably also thank my [ergodox-ez](https://ergodox-ez.com/) learning curve!

proposals/0026/README.md
micahhausler: Fixed

micahhausler:
@displague Fixed cutoffs and did a once-over of the whole thing

proposals/0026/README.md
```diff
@@ -0,0 +1,347 @@
...
+  In order to support a least-privilege access model, several new authorization modes would need to be supported.
+  Implementation of such robust authorizations is non-trivial, and would require extensive initial design work.
+  * An attribute-based authorization method for Tink workers.
+    At the time of writing, any Tink worker is authorized to list all hardware data, including possibly sensitive metadata for other hardware.
+    Today, the Tinkerbell Cluster API (CAPI) provider stores machine CA private key data in Tinkerbell hardware metadata.
+    Ideally, some authorizer that could make decisions based on authentication identity would be used (The Kubernetes analog would be the [Node Authorizer][node-authorizer]).
+    The authentication method in this model could be a network identity, a TPM, or some other bootstrapped authentication credential.
+  * A role-based access method administrative clients.
+  * Either a role-based access method, or some forwarding mechanism for passthrough clients like Boots and Hegel.
+  *
```
displague: stray asterisk

displague (dismissed):
Thanks for the details and revisiting the document for follow-ups, @micahhausler!

proposals/0026/README.md
micahhausler: Fixed

micahhausler:
Updated the status to `Published`

jacobweinstock (approved):

jacobweinstock:
Hey @tstromberg, when/if you have some cycle would you mind checking this out again, please?

displague (approved):

tstromberg (approved):
Let it be merged! Thank you for addressing the bootstrapping concerns.

PR #54, opened 2021-12-14
revise tink, sandbox, and documentation versioning processes

displague:
## Description

Revise tink, sandbox, and documentation versioning processes described in proposal 0002. 
The goal of these changes is to allow for an initial set of 0.1.0+ tags to be added to the existing projects where no tags are currently defined.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

-----
[View rendered proposals/0002/README.md](https://github.com/tinkerbell/proposals/blob/displague-patch-1/proposals/0002/README.md)

### DIFF
--- a/proposals/0002/README.md
+++ b/proposals/0002/README.md
@@ -2,13 +2,12 @@
 id: 0002
 title: Set the mood around releasing code
 status: accepted
-authors: Gianluca Arbezzano <gianarb92@gmail.com>
+authors: Gianluca Arbezzano <gianarb92@gmail.com>, Marques Johansson <mjohansson@equinix.com>
 ---
 
 ## Summary
 
-This proposal sets the expectation about what it means to release code part of
-the Tinkerbell organisation.
+This proposal defines release behaviors for projects in the Tinkerbell organisation.
 
 ## Goals and non-Goals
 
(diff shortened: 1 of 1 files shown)


displague:
In the Dec 14 Triage Party, we closed the existing https://github.com/orgs/tinkerbell/projects projects in (roadmap, repo health, issues list) in favor of new projects based on intended feature sets. Issues lists (a triage system) was closed in favor of triage parties tracking, which is label based. Repo health was a project focused on specific early milestones for providing minimal standards within Tinkerbell projects.

Projects can be seen as feature sets grouping multi-repo efforts.  In the triage party meeting, we also discussed tying this PR's proposal updates to projects.  For example, `v0.8.0` can be seen as an organizational goal, a project.  After `v0.7.0` is released, the timing of which should be a community meeting discussion, we would assemble issues that we would like to focus on for `v0.8.0`.  Assembling issues could take place during the community meeting or asynchronously via issue comments and voting (the rules of which I won't attempt to pin down).  Issues labels may be used to represent these goals in automation, e.g. `tinkerbell/0.7.0` (or `milestone/` or `target/`), if necessary.  

Each project should create milestones representing upcoming semantic revision targets.  If we are on v0.5, a milestone for v0.6, v0.7, v1.0 could be created. Issues could then be targeted for specific milestones and retargeted when missed.

We should try to avoid duplicating purpose and functionality between projects, issues, labels, and milestones. 

I'll revisit this PR and see if any of this thinking should be codified into the proposal update.

displague:
Another point that was discussed was the idea of mapping sandbox release tags to the tags of its dependencies.

If sandbox is at 0.6.0 today, we would tag 0.6.0 on the git commit represented in sandbox 0.6.0 for each project. The appropriate git commit may need to be determined by docker image shas, mapping these to the CI build git sha.

If this could be done for version 0.6 down to 0.1, we will have more history - a better way to talk about and reason about past states of the project.  Our project's generated changelogs will be meaningful.

Doing this may be ambitious, so the community was ok with proceeding with the original idea of stamping 0.1.0 on all untagged projects if we can not easily map 0.6.0 downwards.

displague:
I wanted to capture the versions of tinkerbell/sandbox dependent tools at each version of sandbox - however, it appears that this was never "fixed", sandbox 0.4.0, if deployed today would pull down the same versions used by 0.6.0.

In the future, we should ensure that https://github.com/tinkerbell/sandbox/blob/main/deploy/compose/registry/registry_images.txt refers to fixed shas or tags, not `latest`.  Like https://github.com/tinkerbell/sandbox/blob/456b73874e90e48651d2ffa38c6b5226d3b6d5b7/deploy/compose/.env.

crayzeigh (approved):
lgtm

cprivitere (approved):

PR #57, opened 2022-07-21
Add hardwareRef to workflow spec proposal

srikarsganti:
## Description

Proposal to make hardware data available to workflows using the Kubernetes backend.

## Why is this needed

Cluster API Provider Tinkerbell is used to provision Kubernetes clusters using a Tinkerbell stack. It watches for the TinkerbellMachine custom resource and generates a Workflow and Template custom resource that are used by the Workflow controller for rendering Tink Worker actions. The Workflow Controller currently supports a string-string map used as data for rendering the template.

CAPI enforces a constraint where it uses a single template for all nodes within a group. For example, TinkerbellMachine’s for the control plane are all created using the same template. Often, the Tinkerbell actions defined in a template action override require machine specific information, such as disk, that is unavailable when the workflow is rendered. This forces consumers to specify a single value or leverage an out-of-band mechanism for creating an actions template.

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

-----
[View rendered proposals/0028/README.md](https://github.com/srikarsganti/Proposal/blob/0028/proposals/0028/README.md)


proposals/0028/README.md
```diff
@@ -0,0 +1,35 @@
...
+tasks:
+  - name: foo
+    image: write2disk
+    environment: 
+      DEST_DISK: {{ .hardware.disks[0] }}  
+```
+
+## Future development
+
+Long term we might consider inserting a contract that defines the data and format exposed to templates. This would enable independent evolution of data storage formats and templates. We consider this work out of scope for the initial development.
```
chrisdoherty4:
```suggestion
## Cons

The `Hardware` type is similar to a POJO and represents a storage structure. By exposing it directly to templates we create a data coupling between templates and the data storage structure making it difficult to evolve the storage structure. Given the consumer owns the template and Tinkerbell owns the storage structure its likely we'll want to evolve the storage format that may result in breaking template definitions. 

```

proposals/0028/README.md
```diff
@@ -0,0 +1,35 @@
+---
+id: 0028
+title: Make hardware data available to Kubernetes workflows
+status: discussion
+authors: Srikar Ganti <srikarsganti@gmail.com> Chris Doherty <chris.doherty4@gmail.com>
```
chrisdoherty4:
```suggestion
authors: Srikar Ganti <srikarsganti@gmail.com>, Chris Doherty <chris.doherty4@gmail.com>
```

proposals/0028/README.md
```diff
@@ -0,0 +1,36 @@
...
+  - name: foo
+    image: write2disk
+    environment: 
+      DEST_DISK: {{ .hardware.disks[0] }}  
+```
+
+## Cons
+
+The `Hardware` type is similar to a POJO and represents a storage structure. By exposing it directly to templates we create a data coupling between templates and the data storage structure making it difficult to evolve the storage structure. Given the consumer owns the template and Tinkerbell owns the storage structure its likely we'll want to evolve the storage format that may result in breaking template definitions. 
+
```
chrisdoherty4:
```suggestion
- The `Hardware` type is similar to a POJO and represents a storage structure. By exposing it directly to templates we create a data coupling between templates and the data storage structure making it difficult to evolve the storage structure. Given the consumer owns the template and Tinkerbell owns the storage structure its likely we'll want to evolve the storage format that may result in breaking template definitions. 
- Tinkerbell currently supports actions that run on different Tink Workers. A single `hardwareRef` doesn't cover workflow use cases that leverage multiple Tink Workers. We could address this in the future but it it would need to take a wider view of Tinkerbell components and ensure the commonly observed `device_x` key in the `hardwareMap` isn't integral to identifying hardware resources.

```

proposals/0028/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+```yaml
+apiVersion: "tinkerbell.org/v1alpha1"
+kind: Workflow
+metadata:
+  name: wf1
+  namespace: default
+spec:
+  templateRef: debian
```
chrisdoherty4, line 40:
```suggestion
  templateRef: debian
  hardwareRef:
  	name: hardware-1
  	namespace: hardware-namespace
```

proposals/0028/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Problem
+
+CAPI enforces a constraint where it uses a single template for all nodes within a group. For example, TinkerbellMachine’s for the control plane are all created using the same template. Often, the Tinkerbell actions defined in a template action override require machine specific information, such as disk, that is unavailable when the workflow is rendered. This forces consumers to specify a single value or leverage an out-of-band mechanism for creating an actions template.
+
+## Proposal
+
+We propose adding a hardwareRef field to the [Workflow CRD](https://github.com/tinkerbell/tink/blob/main/pkg/apis/core/v1alpha1/workflow_types.go#L18). CAPT [will be responsible for populating the hardwareRef](https://github.com/tinkerbell/cluster-api-provider-tinkerbell/blob/main/controllers/machine.go#L661) as it acquires the hardware to be used. The Workflow Controller will use the hardwareRef to retrieve the Hardware, serialize it and make it available to the template during rendering. The hardware data will be namespaced under a .hardware key. Defensive code will be added to ensure the hardwareMap does not specify a hardware key that could overwrite the .hardware.
```
chrisdoherty4:
```suggestion
We propose adding a hardwareRef field to the [Workflow CRD](https://github.com/tinkerbell/tink/blob/main/pkg/apis/core/v1alpha1/workflow_types.go#L18). CAPT [will be responsible for populating the hardwareRef](https://github.com/tinkerbell/cluster-api-provider-tinkerbell/blob/main/controllers/machine.go#L661) as it acquires the hardware to be used. The Workflow Controller will use the hardwareRef to retrieve the Hardware, serialize it and make it available to the template during rendering. The hardware data will be namespaced under a .hardware key. The existing `hardwareMap` will remain untouched and defensive code will be added to ensure the hardwareMap does not specify a hardware key that could overwrite the .hardware.
```

jacobweinstock:
Thanks @srikarsganti , I really like this and I'm very excited for this functionality.

jacobweinstock (dismissed):

proposals/0028/README.md
```diff
@@ -0,0 +1,52 @@
...
+  	name: hardware-1
+  	namespace: hardware-namespace
+  hardwareMap:
+    device_1: 3c:ec:ef:4c:4f:54
+```
+
+## Cons
+
+- The `Hardware` type is similar to a POJO and represents a storage structure. By exposing it directly to templates we create a data coupling between templates and the data storage structure making it difficult to evolve the storage structure. Given the consumer owns the template and Tinkerbell owns the storage structure its likely we'll want to evolve the storage format that may result in breaking template definitions. 
+- Tinkerbell currently supports actions that run on different Tink Workers. A single `hardwareRef` doesn't cover workflow use cases that leverage multiple Tink Workers. We could address this in the future but it it would need to take a wider view of Tinkerbell components and ensure the commonly observed `device_x` key in the `hardwareMap` isn't integral to identifying hardware resources.
```
displague, line 51: What complications would be introduced by using `hardwareRefs` (plural)?

chrisdoherty4: It's more complicated than the commentary suggests. Tasks within a workflow are associated with a to worker ID. Workers ask for actions using their ID and tasks in a workflow are expected to specify that ID. The ID is a free-form string defined by the user. Specifying multiple hardware references implies we know how to associate a consumer defined string with a piece of hardware (and we don't, or at least not easily). 

I think this whole thing needs re-thinking and simplifying but this proposal doesn't seek to address that. It only seeks to alleviate a frustrating constraint in template rendering for the typical use case where a workflow runs on a single worker.

proposals/0028/README.md
```diff
@@ -0,0 +1,52 @@
...
+## Example template
+
+This example illustrates the use of disks retrieved from the hardware type used in a template.
+
+```yaml
+tasks:
+  - name: foo
+    image: write2disk
+    environment: 
+      DEST_DISK: {{ .hardware.disks[0] }}
```
displague, line 30: Is hardware defined as immutable in the Hardware spec? Are there concerns about action definitions being raced with hardware changes?

chrisdoherty4: Not particularly. If some value doesn't exist at the time the template is rendered then it doesn't exist (or has changed) and we error out.

chrisdoherty4:
Needs merging.

@displague @jacobweinstock

displague (dismissed):

jacobweinstock (approved):

PR #58, opened 2022-07-22
[0029] Remove Postgres database support

chrisdoherty4:
[View rendered proposals/0029/README.md](https://github.com/chrisdoherty4/tinkerbell-proposals/blob/0029/proposals/0029/README.md)

### DIFF
--- a/proposals/0029/README.md
+++ b/proposals/0029/README.md
@@ -1,7 +1,7 @@
 ---
 id: 0029
 title: Remove Postgres database support
-status: prediscussion
+status: Approved
 authors: Chris Doherty <chris.doherty4@gmail.com>, Jacob Weinstock <jakobweinstock@gmail.com>
 ---
 
--- a/proposals/0029/README.md
+++ b/proposals/0029/README.md
@@ -1,0 +1,20 @@
+---
+id: 0029
+title: Remove Postgres database support
+status: prediscussion
+authors: Chris Doherty <chris.doherty4@gmail.com>, Jacob Weinstock <jakobweinstock@gmail.com>
+---
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


displague:
In the August 16 Community call, we discussed that this proposal would benefit from mentioning the tools/layers affected and unaffected by Postgresql removal.

The proposal should also take into consideration the community (slack) and offer recommendations for those scenarios against the removal. 

It would also be beneficial to explore what if any solutions could be prepared to aid in migrations, and what troubles could be involved in such a solution.

chrisdoherty4:
We've had very little movement from the community on this and those that have commented have welcomed the change. 
 
The proposal alludes to using KinD, K3D or some other light weight cluster solution for deploying Kubernetes and still supporting raw Docker deployment of services. I don't think its necessary to explain specifics beyond that, nor do I know what detail to include.

With respect to migrating, I'm not sure we really need to provide anything. Workflows are ephemeral by design. We could try to supply some tooling to convert hardware and template JSON to Kubernetes custom resources but I suspect users could manually convert without much hassle (again, due to a lack of community engagement its hard to gauge). I suspect we'll need to handle this ad-hoc.

Its been 2 months since the proposal was originally written and circulated. We've observed very few sign up so the new mailing lists so I don't think there's much more circulating to be done. I think its decision time.

thebsdbox:
I'm happy to move forward with this.

chrisdoherty4:
One thing I'll add, which really goes without saying: we need the tooling in-place before we can remove anything. It should be relatively easy, and well documented, to get something up and running with the Kube back-end which isn't the case today.

chrisdoherty4:
Good to merge.

@displague @jacobweinstock

displague (approved):

PR #61, opened 2022-09-22
Configurable default CAPT templates and simplification of template overriding

chrisdoherty4:
[View rendered proposals/0030/README.md](https://github.com/tinkerbell/proposals/pull/61/files?short_path=8dac48f#diff-8dac48f2724ce87b9a0649262b3ab05a0d6b8191750ac90b6c6c5ba0f3078489)


displague:
This sounds good. Could you add a yaml example of how this affects the resources.

proposals/0030/README.md
```diff
@@ -0,0 +1,32 @@
...
+Using a reference field simplifies the CAPT API and enables re-use of templates when provisioning machines.
+
+## Rationale
+
+The default template generated by CAPT makes assumptions on how nodes are to be provisioned. 
+It assumes images reside in an OCI registry, the operator does not wish to adjust cloud-init configuration, nodes are to be bootstrapped using cloud-init and that users wish to use kexec to launch the node bootstrap process. 
+Allowing operators to specify a default gives operators better control over how nodes are provisioned without needing to specify overrides.
+
+Template overrides are specified using a string field on the `TinkerbellMachineTemplate`. 
+The string field makes it difficult to understand the structure of the template and requires operators to understand how it's rendered to be confident the template is correct.
```
displague: Does this suggest there would be a benefit to having both the template reference and the rendered template in the resource?

chrisdoherty4: The rendered template would feature on the `Workflow` status (as it does today). I can't think of a reason to store it on the `TinkerbellMachine`. Although, perhaps I misunderstand because we don't store the rendered template on `TinkerbellMachine`s today either.

To clarify, the template override string field isn't a rendered template, its an unrendered template used to create a new template that is subsequently rendered as part of `Workflow` execution.

chrisdoherty4:
@displague Added some implementation detail.

chrisdoherty4:
Needs a diagram to depict existing relationships within the system.

